PR #2, opened 2024-10-28
NOT READY : Simplified version of filters example

FairlyBet:
This example shows appliance of a single Emboss filter to an image with minimized amount of code and complexity.


vadimb892 (commented):
I checked this example. It works.

examples/minwebgl/filter/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    gl.generate_mipmap( GL::TEXTURE_2D );
+
+    let canvas = gl.canvas().expect( "Canvas should exist" ).dyn_into::< HtmlCanvasElement >().unwrap();
+    canvas.set_width( img.width() );
+    canvas.set_height( img.height() );
+    gl.viewport( 0, 0, img.width() as i32, img.height() as i32 );
+
+    let vertex_source = include_str!( "shaders/main.vert" );
+    let fragment_source = include_str!( "shaders/main.frag" );
+    let program = gl::ProgramFromSources::new( vertex_source, fragment_source )
```
vadimb892, line 50: You compile shader program every time when image is loaded? Argue your solution, when it can be useful.

PR #4, opened 2024-10-31
READY : Raycaster

FairlyBet:
Simple raycasting demo


JekiXD:
Cool. An idea, what if you build a triangle fan from rays and draw that instead? So that the visible space will be uniform in color. Or maybe you know a better way to color the visible space, instead of drawing a lot of ray separately

FairlyBet:
> > Cool. An idea, what if you build a triangle fan from rays and draw that instead? So that the visible space will be uniform in color. Or maybe you know a better way to color the visible space, instead of drawing a lot of ray separately
> 
> Although, probably you don't need it for this example. Maybe would be better to just draw less lines to better see the result of ray casting, because the map is small. Right now it looks like you are trying to paint the vision field of the player with those lines

Yeah, it is easy to adjust amount of rays ( you can try change `ray_count` variable ), I just picked a little too big amount of rays in order to increase resolution of rendering. Probably I should reduce it :)

JekiXD (approved):

projdysvit (approved):

PR #5, opened 2024-11-01
NOT READY : Mingwebgpu

JekiXD:


module/min/minwebgpu/src/error.rs
```diff
@@ -12,6 +12,10 @@ mod private
     CanvasError( #[ from ] CanvasError ),
     #[ error( "Device error :: {0}" ) ]
     DeviceError( #[ from ] DeviceError ),
+    #[ error( "Device error :: {0}" ) ]
+    ContexError( #[ from ] ContextError ),
+    #[ error( "Device error :: {0}" ) ]
+    TextureErrorError( #[ from ] TextureError ),
```
vadimb892, line 18: I think it bad idea to repeat `Error` twice in variant name

module/min/minwebgpu/src/error.rs
```diff
@@ -22,6 +26,20 @@ mod private
     ConfigurationError( String )
```
vadimb892, line 26: Can `ConfigurationError` be decomposed on more concrete types?

module/min/mingl/src/web/model/obj.rs
```diff
@@ -0,0 +1,198 @@
...
+        \x20 • Normals ind.:   {N_Indicies}\n\
+        Bounding Volume:\n\
+        \x20 • Box:\n\
+        \x20    Min: {BoxMin} \n\
+        \x20    Max: {BoxMax} \n\
+        \x20 • Sphere: \n\
+        \x20    Center: {Center} \n\
+        \x20    Radius: {Radius}\n\
+        ",
+        ModelName = self.report.name,
```
vadimb892, line 66: Variable names in Rust use snake_case, when types use PascalCase. Are you use clippy?

module/min/mingl/src/web/model/obj.rs
```diff
@@ -0,0 +1,198 @@
...
+          \x20 • Ambient texture: {TAmbient} \n\
+          \x20 • Diffuse texture: {TDiffuse} \n\
+          \x20 • Specular texture: {TSpecular} \n\
+          \x20 • Normal texture: {TNormal} \n\
+          \x20 • Shininess texture: {TShininess} \n\
+          \x20 • Dissolve texture: {TDissolve} \n\
+          \x20 • Illumination model: {IllumModel} \n\
+          \x20 • Unknown parameters: {Other} \n\
+          ",
+          Name = m.name,
```
vadimb892, line 127: Also wrong case

vadimb892 (changes requested):

module/min/minwebgpu/src/error.rs
JekiXD: No, context.configure just return jsvalue as a error.

module/min/mingl/src/web/model/obj.rs
JekiXD: It's not a variable name, it is a name for a pattern inside of the macro, just to be able to see the position the property goes to inside the string

PR #6, opened 2024-11-01
READY : Warning fix

FairlyBet:
Workspace lints for some reason warn about unsafe code in the [filters/lil-gui.rs](https://github.com/Wandalen/cgtools/blob/master/examples/minwebgl/filters/src/lil_gui.rs) but there's no unsafe code actually.`

### DIFF
--- a/examples/minwebgl/filters/Cargo.toml
+++ b/examples/minwebgl/filters/Cargo.toml
@@ -18,6 +18,3 @@ web-sys = { workspace = true, features = [
   'WebGlTexture',
   'Location',
 ] }
-
-[lints]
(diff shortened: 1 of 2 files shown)


examples/minwebgl/filters/Cargo.toml
```diff
@@ -18,6 +18,3 @@ web-sys = { workspace = true, features = [
   'WebGlTexture',
   'Location',
 ] }
-
```
Wandalen, line 21: that's not solution

Wandalen (changes requested):

PR #7, opened 2024-11-04
NO REVIEW : Docker and make files as build system

bugsweeper:
Docker and make files for for easier build on clients machine.
Docker file modified for building examples in workspace project


JekiXD (commented):
There is a lot of setup code, that you will have to repeat when making docker and make files for other examples. Is it possible to move all of it to workspace level, and leave only what is relevant to the example specifically? I don't really know how to work with make and docker files that well

PR #13, opened 2024-11-06
READY : Video as texture example

projdysvit:


projdysvit:
Please review 😊

module/min/minwebgl/src/texture/video.rs
```diff
@@ -0,0 +1,38 @@
...
+
+  gl.bind_texture( GL::TEXTURE_2D, texture.as_ref() );
+  gl.tex_parameteri( GL::TEXTURE_2D, GL::TEXTURE_WRAP_S, GL::CLAMP_TO_EDGE as i32 );
+  gl.tex_parameteri( GL::TEXTURE_2D, GL::TEXTURE_WRAP_T, GL::CLAMP_TO_EDGE as i32 );
+  gl.tex_parameteri( GL::TEXTURE_2D, GL::TEXTURE_WRAP_R, GL::CLAMP_TO_EDGE as i32 );
+  gl.tex_parameteri( GL::TEXTURE_2D, GL::TEXTURE_MIN_FILTER, GL::LINEAR as i32 );
+  gl.tex_parameteri( GL::TEXTURE_2D, GL::TEXTURE_MAG_FILTER, GL::LINEAR as i32 );
+
+  texture
+}
```
JekiXD, line 20: I think it's better to move this to texture/d2

examples/minwebgl/video_as_texture/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+  video_element.set_height( video_height );
+  video_element.set_loop( true );
+  video_element.set_muted( true );
+  video_element.set_src( path );
+  let _ = video_element.play()?;
+
+  wasm_bindgen_futures::JsFuture::from( load_promise ).await?;
+
+  Ok( video_element )
+}
```
JekiXD, line 101: The promise part is quite confusing. You should leave some comments here.

Also, do you think it's general enough to be moved to mingl/web? Would it be useful as a part of the library? Not saying you should move it, just asking for an opinion

examples/minwebgl/video_as_texture/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+
+  video_element.set_width( video_width );
+  video_element.set_height( video_height );
+  video_element.set_loop( true );
+  video_element.set_muted( true );
+  video_element.set_src( path );
+  let _ = video_element.play()?;
+
+  wasm_bindgen_futures::JsFuture::from( load_promise ).await?;
+
```
JekiXD, line 99: You don't need to include wasm_bindgen_futures, because it's reexported from mingl, you can just do gl::JlFuture, if you enable "future" feature on minwebgl

examples/minwebgl/video_as_texture/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+  let data_buffer = gl::buffer::create( &gl )?;
+  gl::buffer::upload( &gl, &data_buffer, &data, gl::GL::STATIC_DRAW );
+  gl::BufferDescriptor::new::< [ f32; 2 ] >()
+  .stride( 4 )
+  .offset( 0 )
+  .attribute_pointer( &gl, position_attribute_location, &data_buffer )?;
+  gl::BufferDescriptor::new::< [ f32; 2 ] >()
+  .stride( 4 )
+  .offset( 2 )
+  .attribute_pointer( &gl, tex_coord_attribute_location, &data_buffer )?;
```
JekiXD, line 47: This buffer setup is unnecessary. You can define this arrays in the shader, and use glVertexIndex to get the value

JekiXD (changes requested):

examples/minwebgl/video_as_texture/src/main.rs
projdysvit: Yes, it would be great to see that functionality as part of lib :)

examples/minwebgl/video_as_texture/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+      (
+        move ||
+        {
+          // When the oncanplay event is triggered, call the resolve function to complete the Promise.
+          resolve.call0( &gl::wasm_bindgen::JsValue::NULL ).unwrap();
+        }
+      );
+
+      // The oncanplay callback calls resolve to complete the Promise.
+      video_element.set_oncanplay( Some( oncanplay.as_ref().unchecked_ref() ) );
```
JekiXD, line 71: You removed the `reject` part. If an error happens, would this promise forever wait for completion, or is it somehow is going to reject itself?

JekiXD: Also I see you use this structure in another PR. If you think its useful and can be generilized, then it's better to add it to the library. To mingl/web, and then reuse it in minwebgl

projdysvit: Yes, I think the promise will wait indefinitely, and this is a problem, it must be fixed)

projdysvit: > Also I see you use this structure in another PR. If you think its useful and can be generilized, then it's better to add it to the library. To mingl/web, and then reuse it in minwebgl

Soon.

JekiXD (approved):

FairlyBet (approved):

PR #17, opened 2024-11-11
NO REVIEW: Minwebgpu + Deffered rendering example on webgpu

JekiXD:


FairlyBet:
Add a little description to shaders please, like what is the purpose of a shader

examples/minwebgpu/deffered_rendering/src/light.rs
```diff
@@ -0,0 +1,155 @@
...
+    )
+  }  
+}
+
+fn generate_lights() -> Vec< Light >
+{
+  let mut rng = rand::thread_rng();
+  
+  let mut lights = Vec::new();
+  for _i in 0..NUM_LIGHTS
```
FairlyBet, line 128: Just in case, you can also write `for _ in 0..NUM_LIGHTS`

module/min/minwebgpu/src/binding_type/sampler.rs
projdysvit, line 1: 1. Why can't we just use one set_type function?
```rust
/// Sets the type of the sampler to `Filtering`
pub fn filtering( mut self ) -> Self
{
  self.s_type = Some( GpuSamplerBindingType::Filtering );
  self
}
/// Sets the type of the sampler to `NonFiltering`
pub fn non_filtering( mut self ) -> Self
{
  self.s_type = Some( GpuSamplerBindingType::NonFiltering );
  self
}
/// Sets the type of the sampler to `Comparison`
pub fn comparison( mut self ) -> Self
{
  self.s_type = Some( GpuSamplerBindingType::Comparison );
  self
}
```

module/min/minwebgpu/src/binding_type/texture.rs
projdysvit, line 1: * In many case you using `Option< T >`, why not init defaults fields it right away?
* Same - using `&mut self` instead of `mut self`.

projdysvit (commented):
In some cases you are using Option<T>, when you have, for example `bool` default - `false`, why not fill it out right away?

Maybe i missed something, let me know 😊

module/min/minwebgpu/src/binding_type/sampler.rs
JekiXD: These are just shortcuts

JekiXD:
> 1. In some cases you are using Option, when you have, for example `bool` default - `false`, why not fill it out right away?
> 2. Need documentation.
> 
> Maybe i missed something, let me know please 😊

If you were to write it in javascript, there are properties you can omit. Those are marked with Option. They will default to whatever javascript will default them to

PR #18, opened 2024-11-11
READY : Spinning cube size optimization example update

projdysvit:
1. Using lib math instead of glam dep.
2. Using lol_alloc instead of wee_alloc, because the last one [seems unmaintained](https://github.com/rustwasm/wee_alloc/issues/107) and [has a leak](https://github.com/rustwasm/wee_alloc/issues/106).


examples/minwebgl/spinning_cube/src/main.rs
```diff
@@ -63,7 +64,7 @@ fn run() -> Result< (), gl::WebglError >
 
   let projection_matrix_location = gl.get_uniform_location( &program, "projection_matrix" );
   let angle_location = gl.get_uniform_location( &program, "angle" );
-  gl::uniform::matrix_upload( &gl, projection_matrix_location, &projection_matrix.to_cols_array()[ .. ], true ).unwrap();
+  gl::uniform::matrix_upload( &gl, projection_matrix_location, &projection_matrix.to_array()[ .. ], true ).unwrap();
```
FairlyBet, line 67: It is not necessary but you can use `raw_slice` when uploading matrices:
``` rust
gl::uniform::matrix_upload( &gl, projection_matrix_location, projection_matrix.raw_slice(), true ).unwrap();
```

PR #49, opened 2025-04-27
BUG : Math bug

JekiXD:


Wandalen:
@JekiXD add explanation to test itself, please

module/math/ndarray_cg/tests/inc/d2_test/creation.rs
```diff
@@ -0,0 +1,16 @@
+use super::*;
```
Wandalen, line 1: explanations

Wandalen (changes requested):

PR #60, opened 2025-05-27
 READY : Deferred shading

FairlyBet:


examples/minwebgl/deferred_shading/src/main.rs
```diff
@@ -81,76 +62,83 @@ fn run() -> Result< (), gl::WebglError >
...
+        rand::random_range( -110.0..=-90.0 )
+      ),
+      ..EllipticalOrbit::random()
+    }
+  ).collect::< Vec< _ > >();
+  let offsets = ( 0..max_light_count )
+  .map( | _ | rand::random_range( 0.0..=( f32::consts::PI * 2.0 ) ) ).collect::< Vec< _ > >();
+  let mut light_colors = ( 0..max_light_count )
+  .map( | _ | random_rgb_color() ).collect::< Vec< _ > >();
+  light_colors[ 0 ] = [ 0.5, 0.5, 0.5 ];
```
vadimb892, line 125: Why individualy set first light color? Can you add explanation here?

examples/minwebgl/deferred_shading/src/main.rs
```diff
@@ -172,43 +158,95 @@ fn run() -> Result< (), gl::WebglError >
...
+    {
+      fps_counter.set_text_content( Some( &format!( "fps: {}", fps ) ) );
+      fps = 0;
+    }
+    last_time = current_time;
+    fps += 1;
+
+    // update light positions
+    let light_count = *light_count.borrow();
+    // from 1 to light_count because we dont update first light source becaues it is global light
```
vadimb892, line 204: *because

examples/minwebgl/deferred_shading/src/main.rs
```diff
@@ -227,62 +265,196 @@ fn run() -> Result< (), gl::WebglError >
...
+fn tex_storage_2d( gl : &GL, format : u32, width : i32, height : i32 ) -> Option< WebGlTexture >
+{
+  let tex = gl.create_texture()?;
+  gl.bind_texture( GL::TEXTURE_2D, Some( &tex ) );
+  gl.tex_storage_2d( GL::TEXTURE_2D, 1, format, width, height );
+  gl::texture::d2::filter_nearest( gl );
+  Some( tex )
+}
+
+fn wtf( _ : f32 ) {}
```
vadimb892, line 456: Why you need this function?

FairlyBet: For magic ✨

examples/minwebgl/deferred_shading/src/main.rs
FairlyBet: because this is the big static light source, its values are not supposed to be random, i added comments about this

PR #123, opened 2025-09-03
APPROVED : fix tests

vadimb892:
This PR:
- removes features and makes dependencies required, because they used always in crate animation. So feature enabled isn't needed. 
- also I fix test_step_function, because previously I changed Step struct behavior.
- part of tiled_renderer integration tests now used only for Linux because they use Linux utils and failing ion Windows.
- fixed compilation errors in canvas_renderer doc tests of readme.md.


wanguardd (approved):
Awesome stuff on 'NEED REVIEW : fix tests', @vadimb892\! This is clean, follows all our best practices, and is ready to go. Approved. ✅

PR #130, opened 2025-09-22
NEED REVIEW : Replaced render logic with tile map renderer

FairlyBet:


FairlyBet:
Hey @wanguardd, thanks for the review! A few notes on your feedback:

🧹 Cleanup
I removed the dead code you mentioned—those warnings shouldn’t be showing up anymore.

❓ Indentation
I honestly don’t see any 4-space indentation in this branch. Could you point me to the exact file/line where you spotted it? (My guess is this might just be a glitch on your side.)

⚠️ CI
There are no CI checks for this branch, and that’s expected. CI might be added to the repository at some point, but it will be done separately and not as part of this PR.

Thanks again for the thorough pass!

wanguardd (approved):

wanguardd (approved):

PR #183, opened 2026-04-03
NOT READY : Update web_sys and fix compilation errors

FairlyBet:
Fixed few errors caused by updated web_sys. All of them are number that previously were i32 but now web_sys returns f64 so I added cast to i32


wanguardd (changes requested):
System of Review completed analysis of 'Update web_sys and fix compilation errors'. The `web-sys` fixes across `filter`, `object_picking`, `browser_input`, and `texture/d2.rs` are correctly applied and the wasm-bindgen JS API surface for `JewelryRenderer` is well-structured. Found **8 findings** across the new `jewelry_configurator` crate and its renderer integration.

---

## 📋 Review Summary

| Category | Count |
|----------|-------|
| 🐛 CORRECTNESS | 3 |
| 📐 DESIGN | 2 |
| 🔍 QUALITY | 3 |

---

## 🐛 CORRECTNESS

🐛 **Material trait API divergence blocks merge into master.** The error-fix branch uses the old `Material` trait method names (`fn get_id`, `fn get_cull_mode`, `fn get_vertex_shader`, `fn get_fragment_shader`) throughout — in both the `PbrMaterial` renderer implementation and in the two new `jewelry_configurator` materials. Master renamed all of these (to `fn id`, `fn cull_mode`, `fn vertex_shader`, `fn fragment_shader`). A local merge attempt has already produced unresolvable conflicts in 4 renderer files: `material/pbr.rs`, `post_processing/composer.rs`, `shaders/bake.frag`, and `shaders/main.frag`.

**Evidence:**
```
module/helper/renderer/src/webgl/material/pbr.rs  line 772:
<<<<<<< HEAD
    fn cull_mode( &self ) -> Option< CullMode >
=======
    fn get_cull_mode( &self ) -> Option< CullMode >
>>>>>>> error-fix

examples/minwebgl/jewelry_configurator/src/gem_material.rs  line 88:
    fn get_cull_mode( &self ) -> Option< CullMode >   // old name
    fn get_id( &self ) -> Uuid                        // old name
    fn get_vertex_shader( &self ) -> String           // old name
    fn get_fragment_shader( &self ) -> String         // old name

examples/minwebgl/jewelry_configurator/src/surface_material.rs  (same pattern)
```

**Action:**
- [ ] Rebase error-fix onto the current master and resolve the 4 merge conflicts in the renderer
- [ ] Rename all `Material` implementations in `GemMaterial` and `SurfaceMaterial` to the new method names: `fn id`, `fn cull_mode`, `fn vertex_shader`, `fn fragment_shader`

---

🐛 **`GemMaterial` and `SurfaceMaterial` never clear the dirty flag — uniforms re-uploaded every frame.** Both materials declare `pub needs_update: bool` (a plain `bool`, not `Cell<bool>` as in `PbrMaterial`) and do not override `clear_needs_update()`. The renderer calls `material.clear_needs_update()` after uploading uniforms (renderer.rs lines 895, 965), but the trait's default implementation is a no-op:

```rust
// Material trait in error-fix branch (material/mod.rs line 141):
fn clear_needs_update( &self ) {}   // default no-op

// GemMaterial: no override, needs_update starts as `true` in new()
// Result: clear_needs_update() is always a no-op for GemMaterial/SurfaceMaterial
//         needs_update() always returns true → configure() called every frame
```

**Evidence:**
```rust
// gem_material.rs line 56
pub needs_update : bool,
// gem_material.rs constructor — starts true, never cleared:
needs_update : true,

// renderer.rs line 892:
if material.needs_update() || last_material_id != Some( material.get_id() )
{
  material.clear_needs_update();  // no-op for GemMaterial — flag stays true
  // ... configure() called every frame
```

**Action:**
- [ ] Add `fn clear_needs_update( &self ) { self.needs_update = false; }` to both `GemMaterial` and `SurfaceMaterial`, OR change the field to `std::cell::Cell<bool>` and override with `self.needs_update.set(false)` to match the pattern used in `PbrMaterial`

---

🐛 **Four shaders missing `#version 300 es` while using ES 3.00 features.** `gem.frag`, `gem.vert`, `surface.frag`, and `surface.vert` all use WebGL2/GLSL ES 3.00 constructs (`layout(location = ...)`, `in`/`out` interpolation qualifiers, multiple render targets) without the required `#version 300 es` directive. Without it, the default is GLSL ES 1.00 and the shader is technically invalid.

**Evidence:**
```glsl
// gem.vert — first 3 lines:
// Generated with Shader Minifier 1.5.1
precision highp float;          // no #version 300 es
layout(location=0)in vec3 V;   // ES 3.00 feature without version directive

// surface.frag — first 3 lines:
precision mediump float;        // no #version 300 es
// ...
layout( location = 0 ) out vec4 frag_color;     // ES 3.00 MRT output
layout( location = 1 ) out vec4 emissive_color; // ES 3.00 MRT output

// gen_cube_map.frag (for comparison — correct):
#version 300 es
precision mediump float;
```

**Action:**
- [ ] Add `#version 300 es` as the very first line (before any other content including comments) to `gem.frag`, `gem.vert`, `surface.frag`, and `surface.vert`

---

## 📐 DESIGN

📐 **`gem.frag` and `gem.vert` are committed as Shader Minifier output without the original source.** Both files open with `// Generated with Shader Minifier 1.5.1` and use single-letter variable and function names (`V`, `B`, `z`, `m`, `d`, `l`, etc.). This makes the shaders impossible to debug or modify without regenerating from the original. The original unminified source must live alongside the generated output.

**Evidence:**
```glsl
// gem.vert lines 1–4:
// Generated with Shader Minifier 1.5.1 (https://github.com/laurentlb/Shader_Minifier/)
precision highp float;
layout(location=0)in vec3 V;
layout(location=1)in vec3 B;
layout(location=2)in vec2 z;
// ... all identifiers are single letters
```

**Action:**
- [ ] Add the unminified `gem.frag` and `gem.vert` source files alongside the minified versions (e.g., `gem.frag.glsl` / `gem.vert.glsl`)
- [ ] Document in `readme.md` how to regenerate the minified versions from source

---

📐 **`[profile.release]` in a workspace member `Cargo.toml` is silently ignored by Cargo.** The `jewelry_configurator/Cargo.toml` contains a `[profile.release]` section with `opt-level = 's'` and `lto = true`, but Cargo only reads profile settings from the workspace root manifest. Profile sections in member manifests have no effect.

**Evidence:**
```toml
# examples/minwebgl/jewelry_configurator/Cargo.toml
[profile.release]
opt-level = 's'
lto = true          # silently ignored — Cargo reads profiles from workspace root only
```

**Action:**
- [ ] Move the desired release profile settings to the workspace root `Cargo.toml` under `[profile.release]` (or a named profile), or document in a comment that this is intentionally a template for the root to copy

---

## 🔍 QUALITY

🔍 **`camera_locked` field is written but never read.** The field is set to `true` in `set_camera_internal()` and back to `false` when orbit controls are re-enabled, but no code path reads it in a conditional. The locking state is already implicitly encoded in `controls.rotation.enabled` and `controls.zoom.enabled`, making `camera_locked` redundant dead state.

**Evidence:**
```rust
// lib.rs — all uses of camera_locked:
348:  camera_locked : bool,        // field declaration
377:      camera_locked : false,   // constructor
450:    self.camera_locked = true;  // set in set_camera_internal()
616:    self.camera_locked = false; // reset when controls re-enabled
// No `if self.camera_locked` or equivalent read anywhere in the file
```

**Action:**
- [ ] Remove `camera_locked` from the struct, or implement a read path that uses it (e.g., guard pointer/touch event handlers against camera movement during a locked animation)

---

🔍 **`downloadScreenshot()` does not restore `bloom_strength` on exception, and the background-detection heuristic is fragile.** The function zeros `bloom_strength` before capturing and restores it at the end, but there is no `try/finally` guard. If any canvas operation throws between those two assignments, bloom stays at `0` permanently. Additionally, sampling pixel `[0, 0]` to identify the background color assumes the top-left corner is always background — this breaks when jewelry or UI elements touch that corner.

**Evidence:**
```javascript
// animation.js lines 45–67 (simplified):
const savedBloom = params.bloom_strength;
params.bloom_strength = 0;                      // bloom disabled
await new Promise( r => requestAnimationFrame(...) );

// ... canvas operations that could throw ...

const bgR = data[0], bgG = data[1], bgB = data[2]; // pixel [0,0] as background
// ... erase matching pixels ...

params.bloom_strength = savedBloom;             // only restored if no throw
// no try/finally around this
```

**Action:**
- [ ] Wrap the function body in `try { ... } finally { params.bloom_strength = savedBloom; }` to guarantee restoration
- [ ] Consider a more reliable background-detection strategy (e.g., a CSS `background-color` declaration or a transparent canvas with compositing)

---

🔍 **Typo `trasnparentA` used consistently in `surface.frag`.** The output variable at location 2 is spelled `trasnparentA` (transposed 's' and 'n') on both the declaration and the write site. While consistent so it compiles, the typo will be confusing to anyone reading the shader.

**Evidence:**
```glsl
// surface.frag lines 12, 30:
layout( location = 2 ) out vec4 trasnparentA;   // ← typo
// ...
trasnparentA = vec4( finalColor * a_weight, alpha );  // ← typo propagated
```

**Action:**
- [ ] Rename `trasnparentA` → `transparentA` in both the declaration and the write site

---

>
>

PR #191, opened 2026-05-19
APPROVED : Retained-mode tilemap_scene Scene + Renderer with delta cache and sprite batching

FairlyBet:
## Summary

Migrates `tilemap_scene` from the stateless `compile_assets` + `compile_frame` API to a retained-mode pipeline (Path A):

- `Scene` owns the rendering world — typed `InstanceHandle`s, per-instance overrides, spatial indexes. Mutated via `spawn` / `despawn` / `move_to` / `set_state` / `set_visible` / `set_tint` / `set_phase_offset` / `set_external_sprite` / `set_global_tint` / `set_seed`. Exposes a monotonic `revision()` counter.
- `Scene::tick(dt)` advances the master clock and returns `Vec<SceneEvent>` — fires `AnimationCompleted` exactly once per OneShot duration crossing per leaf animation layer.
- `Renderer` is the algorithm half — allocates GPU sprite batches lazily, keeps an idle-replay cache keyed by `(scene.revision, scene.clock, camera fingerprint)`, and emits one instanced `DrawBatch` per `(bucket, sheet, blend, clip)` group. Single-key sorted buckets (every sprite shares one `(sheet, blend, clip)`) batch through a single draw call whose instance-buffer order equals the sort order; multi-key sorted buckets keep the per-sprite fallback.

Closes the high-impact items raised by the `render_adapter` integration team: no more runtime spec mutation (per-instance phase/tint overrides), assets compiled once at `Renderer::new`, per-instance `Placement`, interned `InstanceHandle` / `ObjectHandle` u32 ids replacing `Vec<String>` on the hot path, OneShot completion events, and an idle-replay cache that closes "compile_frame rebuilds the whole command stream every frame".

A follow-on batch of fixes / features driven by a first downstream consumer's integration report is bundled below — small enough that splitting into a separate PR would be churn, big enough to call out separately.

## Commits

Retained-mode migration (initial):

1. `chore(tilemap_scene): add slotmap dependency` — workspace + crate Cargo.toml.
2. `feat(tilemap_scene): retained-mode Scene + Renderer with delta cache and sprite batching` — all source, doc, and migrated-compile-test changes.
3. `test(tilemap_scene): integration tests for retained Scene, events, Renderer, and idle-replay cache` — four new integration test files.

Follow-on from downstream consumer feedback:

4. `fix(tilemap_scene): onshot starting not at 0 time` — preliminary OneShot timing fix (master clock → per-instance origin).
5. `fix(tilemap_scene): restart OneShot on set_state` — adds `Instance.state_entered_time`; `set_state` now restarts OneShot from frame 0 and re-arms `AnimationCompleted`. Unblocks attack / death / pulse animations on long-lived instances.
6. `feat(tilemap_scene): typed object/state Catalog with fail-fast validation` — `Scene::catalog().require_object(...).require_state(...).build()?` resolves all required ids at once; `Catalog::object` / `Catalog::state` are panic-on-miss after a successful build, so the hot reconcile path is `.expect()`-free.
7. `feat(tilemap_scene): Scene::tick_into for caller-owned event buffers` — sibling to `tick(dt)`; zero-allocation `tick_into(dt, &mut Vec<SceneEvent>)`.
8. `fix(tilemap_scene): validate atlas frame bounds at compile time` — adds optional `image_size : Option<(u32, u32)>` on `AssetKind::Atlas`; `compile_assets` raises `CompileError::FrameOutOfBounds` when a resolved frame rect lies past the declared image. Catches the "frame 64 silently samples transparent pixels at row 8" failure mode.
9. `docs(tilemap_scene): clarify Atlas.frames vs frame_rects and HashCoord fallback` — names `columns` as the **addressable** column count of the auto-numbering scheme; `frames` as the named-cell manifest; `frame_rects` as explicit pixel rectangles. Documents `PhaseOffset::HashCoord` degenerate behaviour on non-hex placements.
10. `feat(tilemap_scene): PhaseOffset::Instance for placement-agnostic stagger` — new variant; derives per-instance phase from a seed stamped at `spawn` time, mixed through `hash_coord`'s avalanche with the animation id. Works on every placement variant (unlike `HashCoord`, which degenerates on `FreePos` / `Viewport`).
11. `docs(tilemap_scene): roadmap backlog from downstream feedback` — captures the three deferred items (animated `move_to`, OneShot frame easing, image-aware atlas validation as default-on) plus generalisations worth tracking.
12. `docs(tilemap_scene): drop references to untracked feedback docs` — strip pointer-references to local-only working documents from the new code / doc comments.

Second follow-on batch (sorted batching + cache & adapter polish):

13. `feat(tilemap_scene): sorted batching, Set elision, cleanup, hex helper` — single-key sorted buckets collapse into one `DrawBatch` (instance-buffer order = sort order); `emit_or_update_batch` runs a bit-equal pre-pass and emits `SetSpriteInstance` only for slots whose payload actually changed (idle ticks where only `Scene::clock` advances no longer re-upload the GPU instance buffer); `Renderer::cleanup() -> Vec<RenderCommand>` drains live batches into `DeleteBatch` commands so consumers cycling renderers within one backend context can release GPU state on demand; `HexConfig::from_hex_size(w, h, tiling)` computes equilateral-hex `grid_stride` so authors with a sprite bounding box can skip the ¾ formula. Also fixes a latent ordering bug where all `DrawBatch`es were deferred to the end of `cmd_buf` after per-sprite `Sprite`s — now emitted inline at each bucket's pipeline position.
14. `feat(tilemap_renderer): dispatch ScreenSpaceSprite through SVG sprite path` — SVG previously dropped `ScreenSpaceSprite` as a no-op stub; route it through `cmd_sprite` since the SVG pipeline already operates in screen-space coordinates and the payload is identical to `Sprite`.
15. `docs(tilemap_scene): mark sorted batching, cleanup, SVG, hex helper shipped` — roadmap update; raises baseline.
16. `docs(tilemap_scene,tilemap_renderer): sync readmes with shipped work` — `tilemap_scene/readme.md` directory-layout entry now describes the actual batching policy (any single-key bucket batches; sorted single-key buckets preserve order via the instance buffer); `tilemap_renderer/readme.md` "known issues" entry on `ScreenSpaceSprite` narrows to the terminal adapter only.

## Architecture

```
┌──────────────────────────┐         ┌───────────────────────────┐
│ Consumer (game / adapter)│         │ Backend (WebGL / SVG / …) │
└────────────┬─────────────┘         └─────────────▲─────────────┘
             │ spawn / despawn /                  │ submit(commands)
             │ move_to / set_state / tick(dt)     │
             ▼                                    │
   ┌──────────────────┐   &scene, &camera  ┌─────────────────────┐
   │      Scene       │ ──────────────────▶│      Renderer       │
   │ (retained world) │                    │ (algorithm + cache) │
   └──────────────────┘                    └─────────────────────┘
             │                                    │
             └────────── Arc<RenderSpec> ─────────┘
                         (immutable, shared)
```

## Test plan

- [x] `cargo build -p tilemap_scene` clean.
- [x] `cargo test -p tilemap_scene` — 167 tests pass across 9 binaries:
  - 38 unit (hash, ids, camera, coords, animation incl. OneShot origin + PhaseOffset::Instance, neighbors, conditions, vertex, edges, viewport, snapshot expansion).
  - 17 `scene_state_test` — handle resolution, mutation API, `revision()` semantics.
  - 16 `scene_events_test` — `tick` / `tick_into`, OneShot completion edge cases, OneShot restart on `set_state`, `PhaseOffset::Instance` stagger on FreePos placements.
  - 6 `renderer_test` — per-instance overrides, snapshot bridge round-trip.
  - 18 `renderer_cache_test` — idle replay, mutation / tick / camera invalidation, byte-equality on replay, **Set elision on unchanged batches**, **`cleanup()` drain semantics**.
  - 7 `sorted_batching_test` — single-key sorted batching, multi-sheet fallback, inline `DrawBatch` pipeline order.
  - 3 `hex_config_test` — `from_hex_size` arithmetic for flat-top / pointy-top / square fallback.
  - 42 `scene_model_compile_test` — migrated via `tests/common::flatten_to_sprites`; all existing anchor × source coverage preserved; new atlas frame-bounds tests.
  - 14 `scene_model_test` — RON + serde round-trip.
  - 6 `catalog_test` — `Scene::catalog()` builder happy path, missing-object collection, missing-state-on-known-object, double-report suppression, `try_*` accessors, panic-on-unrequired.
- [x] `cargo test -p tilemap_renderer` — existing tests plus one new SVG unit test (`screen_space_sprite_renders_through_sprite_path`).
- [x] Clippy clean for new code; pre-existing warning baseline preserved.
- [x] Existing `RenderCommand::Sprite` assertions in fixtures kept working via `flatten_to_sprites` / `BatchFlattener` test helpers in `tests/common/mod.rs`.

## Deferred (tracked in `roadmap.md`)

From the original Path-A migration:

1. **Stable per-(instance, layer, emit) slot identity** — bit-equal Set elision already eliminates redundant uploads for unchanged batches, but a `move_to` whose anchor cell changes can still shuffle later slots in the per-anchor iteration. Picking up the `EmitKey → (BatchKey, slot)` map originally sketched in this item closes the residual case.
2. **Multi-key sorted-bucket batching** — single-key sorted buckets now batch; multi-key sorted buckets still emit per-sprite because `DrawBatch` draws the whole buffer and run-splitting would defeat batch reuse. The encoded-depth-key path remains the next move when this matters.
3. Viewport-pass batching — `ScreenSpaceSprite` is still per-sprite.
4. Composite OneShot detection (`Variant` / `NeighborBitmask` / `ViewportTiled`) in `Scene::tick`.
5. ~~Explicit `Renderer::cleanup() -> Vec<RenderCommand>`.~~ **Shipped.**
6. Per-handle dirty tracking in `Scene` (current `revision()` is coarse).
7. `Anchor::Multihex` rendering — still rejected with `UnsupportedAnchor`.
8. ~~`Sprite::PartialEq` upstream in `tilemap_renderer`.~~ **Worked around locally** — `renderer.rs` carries a private `sprite_instance_eq` bit-equal compare (transform / sprite id / tint via `f32::to_bits`).

Polish:

- ~~SVG adapter `ScreenSpaceSprite` dispatch.~~ **Shipped.**
- ~~`HexConfig::from_hex_size` bounding-box helper.~~ **Shipped (constructor).** Explicit stride override field for stylised pixel-art hexes still open (low urgency).

From downstream feedback ("Backlog from downstream consumer feedback" in `roadmap.md`):

9. Interpolated `move_to(handle, dest, duration, easing)` + `SceneEvent::MoveCompleted` — current `move_to` is a teleport.
10. `AnimationTiming::Easing { duration, easing : EasingFn, frames }` for natural spawn / settle curves without hand-authoring per-frame durations.
11. Image-dimension-aware atlas validation as default-on — either extend `AssetResolver` to return image dimensions or require `image_size` on every authored atlas; move the check into `validate.rs`.

## Breaking changes

From the original migration:

- `compile_frame` / `compile_assets` removed from the public API — go through `Renderer::new` + `Renderer::render(&scene, &camera)`.
- `Scene` is now the retained runtime type; the old data struct is renamed to `SceneSnapshot` (still RON-deserialisable, bridged via `Scene::from_snapshot`).
- `Scene::advance_clock` removed in favour of `Scene::tick(dt) -> Vec<SceneEvent>`.
- `Tile.objects: Vec<String>` is no longer the runtime placement API — consumers spawn instances by handle.

From the follow-on batches:

- `AssetKind::Atlas` gains an `image_size : Option<(u32, u32)>` field. RON-deserialised specs are unaffected (`#[serde(default)]`). Rust callers that construct `AssetKind::Atlas { ... }` as a struct literal must add `image_size: None` (or `Some((w, h))` to opt into the bounds check).
- `Instance` gains `state_entered_time : f32` and `instance_phase_seed : u32` public fields. Callers reading `Instance` via `Scene::instance(h)` are unaffected; only direct struct-literal construction breaks (and `Instance` is normally created only inside `Scene::spawn`).
- `Scene::set_state` now resets the OneShot clock on the target instance. Any caller using `set_phase_offset(h, Some(-scene.clock()))` as a manual re-arm workaround after `set_state` must drop the workaround — leaving it in causes a double phase shift.
- `compile::animation::resolve_animation_frame` and `compile::animation::declared_phase_seconds` gain a trailing `instance_seed : Option<u32>` parameter. Direct callers (rare — these are mostly internal) pass `None` to keep prior behaviour for `None` / `Fixed` / `HashCoord` / `Linear` `PhaseOffset` variants.
- `PhaseOffset` gains an `Instance` variant. The enum is `#[non_exhaustive]`, so existing exhaustive matches with a wildcard arm keep compiling.
- The SVG adapter now renders `ScreenSpaceSprite` (previously a no-op). Backends that fed both `Sprite` and `ScreenSpaceSprite` and expected SVG to silently drop the latter will start seeing it in the output. Re-author callers that depended on the drop.
- `Renderer::cleanup` is `#[must_use]`; ignoring the returned stream leaks GPU batches in the backend.


wanguardd (changes requested):
System of Review completed analysis of 'Retained-mode tilemap_scene Scene + Renderer with delta cache and sprite batching'. Found **5 findings** across 3 files.

## 📋 Review Summary

| 📖 DOCUMENTATION | 2 | 2 | 4 |
| 🔍 QUALITY | 0 | 1 | 1 |
| 📋 REQUIREMENTS | 0 | 1 | 1 |
| **Total** | **1** | **4** | **5** |

---

## 📖 DOCUMENTATION

**🔴 Blocking INTRODUCED** — `event.rs:26-31` documents that calling `set_state` does **NOT** re-arm the `AnimationCompleted` event and suggests `set_phase_offset` as the re-trigger workaround. This is incorrect: `scene.rs` `set_state()` resets `state_entered_time` to the current clock, which does re-arm OneShot completion detection. A test named `state_switch_re_arms_oneshot` confirms and documents the actual behavior. Callers who rely on this doc comment will implement the wrong re-trigger pattern — or will incorrectly assume that animation events cannot be replayed on long-lived instances.

**Evidence:**
```rust
// event.rs:26-29 says:
/// Fires exactly once per crossing. Re-entering the state — even
/// via an explicit [`crate::scene::Scene::set_state`] call — does
/// NOT re-arm the event; consumers that want a re-trigger reset
/// the per-instance phase via `Scene::set_phase_offset(...)` so the
/// animation's local time wraps back to zero.

// Contradicted by scene.rs:472 inside set_state():
inst.state_entered_time = self.clock;

// And by scene_events_test.rs:411-438 fn state_switch_re_arms_oneshot()
// with comment: "set_state resets state_entered_time on the instance,
// so re-entering a OneShot state restarts its animation from frame 0
// and re-arms its AnimationCompleted event."
```

**Action:**
- [ ] Update `event.rs:26-31` to document the correct behavior. Replace "does NOT re-arm the event; consumers that want a re-trigger reset the per-instance phase via `set_phase_offset(...)`" with: "**DOES** re-arm the event — `set_state` resets `state_entered_time` to the current clock, restarting the OneShot from frame 0. Calling `set_state` is the correct and sufficient way to re-trigger attack / death / pulse animations on long-lived instances."

---

**🟠 Non-blocking INTRODUCED** — `renderer.rs:64-66` documents `CameraSignature` as "`-0.0 == 0.0` and `NaN != NaN`." Both claims are incorrect for `to_bits()` comparison: `f32::neg_zero().to_bits()` is `0x80000000u32` — distinct from `f32::zero().to_bits()` (`0x00000000u32`) — so `-0.0` and `+0.0` produce **different** cache entries, not the same one. Two `NaN` values that share the same bit pattern (the common case for canonical quiet NaN) produce **equal** signatures, not different ones. The doc claims the opposite of the implementation's actual behavior for the `-0.0` case, and there is an additional inconsistency: `renderer.rs:144-145` (in `sprite_instance_eq`) correctly states "`+0.0 != -0.0`", directly contradicting the `CameraSignature` doc.

**Evidence:**
```rust
// renderer.rs:64-66 (CameraSignature struct doc):
/// `f32` fields are compared via their `to_bits` representation so
/// `-0.0 == 0.0` and `NaN != NaN` (matches what bit-equal replay
/// expects — any NaN treated as "different camera").

// renderer.rs:144-145 (sprite_instance_eq, same file — contradicts above):
/// compared via `to_bits` so the result depends only on the bit
/// pattern (correctly: `+0.0 != -0.0`, any `NaN != NaN`).

// Actual to_bits behavior:
// f32::NEG_ZERO.to_bits() == 0x80000000 != 0x00000000 == f32::ZERO.to_bits()
// Two canonical qNaN values share the same bit pattern → compare equal
```

**Action:**
- [ ] Fix `renderer.rs:64-66` to accurately describe the bit-exact semantics. Replace "`-0.0 == 0.0` and `NaN != NaN`" with "`-0.0 != +0.0` (sign bit differs) and NaN == NaN when both share the same bit pattern." Optionally add a note that camera implementations producing `-0.0` coordinates (e.g. from negating zero velocity) will cause spurious cache misses.

---

**🟡 Non-blocking INTRODUCED** — `renderer.rs:399` contains the inline comment "Group consecutive same-(sheet, blend, clip) sprites" at the `group_sprites` call site, but the function being called explicitly documents non-consecutive merging: "Two non-adjacent runs with the same key end up in the same returned group (sprites are appended to [the existing entry])." The call-site comment "consecutive" directly contradicts the function's own doc and its implementation, which uses an `index_of` HashMap to merge all same-key sprites regardless of their positions in the stream.

**Evidence:**
```rust
// renderer.rs:399 (call site):
// Group consecutive same-(sheet, blend, clip) sprites. Order

// renderer.rs:517-521 (group_sprites function doc — same file):
/// Group a bucket's pre-sorted sprite stream by batch key while
/// preserving original order. Two non-adjacent runs with the same
/// key end up in the same returned group (sprites are appended to
```

**Action:**
- [ ] Update the call-site comment at `renderer.rs:399`. Replace "Group consecutive same-(sheet, blend, clip) sprites" with "Group all same-(sheet, blend, clip) sprites (including non-adjacent runs)" to match the function's documented and implemented behavior.

---

**🟡 Non-blocking INTRODUCED** — `HexConfig::from_hex_size` silently accepts `Square4` and `Square8` tiling strategies and returns a `HexConfig` with `(w, h)` stride. The only indication that these values are invalid is an internal comment "rejected by validation" — invisible to a caller using the public API directly without going through `spec.load()`. A game developer calling `HexConfig::from_hex_size(w, h, TilingStrategy::Square4)` receives a valid struct with no warning, then encounters a runtime error only at scene-load time.

**Evidence:**
```rust
// pipeline.rs:116-124 (from_hex_size):
let grid_stride = match tiling
{
  TilingStrategy::HexFlatTop   => ( w * 3 / 4, h ),
  TilingStrategy::HexPointyTop => ( w, h * 3 / 4 ),
  // Square tilings are rejected by validation; we still produce a
  // sane default (1:1 with the bounding box) rather than panic so
  // callers exploring the API don't trip a destructive failure.
  TilingStrategy::Square4 | TilingStrategy::Square8 => ( w, h ),
};
```

**Action:**
- [ ] Add a `#[doc]` note to `HexConfig::from_hex_size` stating: "Note: `Square4` and `Square8` are reserved and not yet implemented — this function returns a valid `HexConfig`, but `RenderSpec::load()` will reject the scene with `ValidationError::UnsupportedTiling`." Alternatively, document this on the `TilingStrategy` variants themselves.

---

## 🔍 QUALITY

**🟡 Non-blocking INTRODUCED** — `renderer.rs` contains three inline comments referencing internal planning step numbers that are not tracked in this repository: "Step 4.", "Step 4b.", and "Step 4 (Polish #7". These labels are opaque to any reader who was not the original author during the planning phase and do not convey architectural rationale.

**Evidence:**
```
renderer.rs:190: // Per-renderer delta cache — Step 4.
renderer.rs:215: // Sprite-batch state — Step 4b.
renderer.rs:435: // Viewport sprites — always per-sprite for Step 4 (Polish #7
```

**Action:**
- [ ] Replace the three "Step 4 / Polish #7" references in `renderer.rs` with brief descriptive comments explaining the design intent (e.g. why viewport sprites are always per-sprite, why the delta cache is a struct field rather than a local).

---

## 📋 REQUIREMENTS

**🟠 Non-blocking INTRODUCED** — `validate.rs` contains 15 `// TODO SPEC §16` markers for validation rules that are defined as error variants in `error.rs` but not enforced at runtime. Specifically, `MissingDefaultState` and `ReservedId` are fully defined in `error.rs` with all required fields but are never constructed anywhere in `validate.rs`. A scene that declares `default_state: "attack"` without an `"attack"` entry in its `states` map, or that uses the reserved id `"void"` as a user object id, passes validation silently. The module doc acknowledges partial implementation, but the gap is not surfaced in a way that callers can programmatically query.

**Evidence:**
```
validate.rs:144: // TODO SPEC §16: every Asset.id / Tint.id / Animation.id / Effect.id / Object.id unique within its kind.
validate.rs:145: // TODO SPEC §16: every TintRef / AnimationRef / EffectRef resolves.
validate.rs:147: // TODO SPEC §16: for each object — default_state exists in states.
validate.rs:148: // TODO SPEC §16: anchor ↔ sprite-source compatibility (SPEC §3, §5).
validate.rs:150: // TODO SPEC §16: reserved id "void" not declared as user object.
// (15 TODO markers total)
// error.rs defines: MissingDefaultState { object: String, state: String }
// error.rs defines: ReservedId { id: String }
// Neither is constructed in validate.rs (0 construction sites)
```

**Action:**
- [ ] Either implement the missing validation checks before merging, or add a prominently placed module-level doc comment to `validate.rs` explicitly listing which SPEC §16 rules are currently enforced versus deferred, so integrators know the partial validation surface. Tracking them as open issues in `roadmap.md` (which already exists) would also satisfy this requirement.

---

This is a well-structured contribution with a clear architecture. The retained-mode model is internally consistent — revision/clock separation, generational handles, idle-replay cache, and the Catalog builder pattern all fit together cleanly. The test suite is thorough (172 functions across 9 files). The findings above are straightforward to address. Looking forward to the next iteration.

---

>
>

FairlyBet:
@wanguardd 

# Review Response

## 1. Blocking INTRODUCED — `event.rs:26-31` stale doc comment about `set_state` re-arm

**Verdict:** Valid finding. Fixed.

The doc comment in `event.rs` was written before commit `4edae2a2 fix(tilemap_scene): restart OneShot on set_state` which changed the behavior. After that fix, `set_state()` resets `state_entered_time = self.clock` (scene.rs:472), which does re-arm OneShot completion. The doc was never updated.

Also found and fixed an analogous stale reference in `scene_events_test.rs:9` module doc which still said "no-re-arm-on-`set_state` rule".

**Commit:** `982d079b fix(tilemap_scene): correct stale doc — set_state does re-arm OneShot event`

## 2. Non-blocking INTRODUCED — `renderer.rs:64-66` incorrect `to_bits` doc

**Verdict:** Valid finding. Fixed, plus found analogous issue.

The `CameraSignature` doc claimed `-0.0 == 0.0` and `NaN != NaN` under `to_bits()` — both wrong. `f32::NEG_ZERO.to_bits()` is `0x80000000` (differs from `0x00000000`), and canonical quiet NaN shares one bit pattern so compares equal.

Also fixed the same NaN claim in `sprite_instance_eq` doc (line 145) which said "any `NaN != NaN`" — inconsistent with `to_bits` semantics. The `-0.0 != +0.0` part there was already correct.

**Commit:** `cc75d9da fix(tilemap_scene): correct to_bits doc — -0.0 != +0.0 and identical NaN compares equal`

## 3. Non-blocking INTRODUCED — `renderer.rs:399` "consecutive" comment vs non-consecutive merging

**Verdict:** Valid finding. Fixed.

The call-site comment said "Group consecutive same-(sheet, blend, clip) sprites" but `group_sprites` uses a `HashMap<BatchKey, usize>` to merge all same-key sprites regardless of adjacency. Changed "consecutive" → "all".

**Commit:** `a114503a fix(tilemap_scene): call-site comment — group_sprites merges all runs, not only consecutive`

## 4. Non-blocking INTRODUCED — `HexConfig::from_hex_size` silently accepts Square tilings

**Verdict:** Valid finding. Fixed.

`TilingStrategy` enum doc already states Square variants are reserved/rejected at load time (lines 132-134), but `from_hex_size` doc didn't mention this. Added a note that Square4/Square8 are accepted without panic but will be rejected at validation. Not adding a runtime warning/error here — the non-panicking fallback is intentional for API exploration.

**Commit:** `267f6713 docs(tilemap_scene): note that from_hex_size accepts but does not implement Square tilings`

## 5. Non-blocking INTRODUCED — `renderer.rs` opaque "Step 4 / Polish #7" planning labels

**Verdict:** Valid finding. Fixed.

Removed "Step 4.", "Step 4b.", and "Step 4 (Polish #7" from the three comments. The section headers at lines 190 and 215 already had thorough descriptive blocks below — the labels were pure noise. Line 435 was rephrased to keep the useful "not yet implemented" context without the opaque reference.

Also checked for similar patterns in `src/` — found "Slice N" references in `compile/` but those convey scope-limitation semantics ("Slice 1 supports Atlas only"), not arbitrary planning step numbers — left them.

**Commit:** `f9399837 fix(tilemap_scene): remove opaque Step/Polish planning labels from renderer comments`

## 6. Non-blocking INTRODUCED — `validate.rs` unenforced `MissingDefaultState` and `ReservedId`

**Verdict:** Valid finding. Fixed differently than suggested.

Instead of adding more documentation about what's not enforced (the module/trait/impl docs already cover this clearly), implemented the two trivial checks:
- `MissingDefaultState`: `object.states.contains_key(&object.default_state)` — 7 lines
- `ReservedId`: `object.id == "void"` — 4 lines

Removed the corresponding TODO markers; the remaining 6 TODOs stay as-is (they require more design work). Updated the `impl Validate for RenderSpec` doc to list the two new enforced rules.

**Commit:** `3bbd7c48 feat(tilemap_scene): enforce MissingDefaultState and ReservedId validation rules`

wanguardd (changes requested):
## 📊 Review Summary

**Review #2**

**Progress since Review #1**:
- Previous review: 5 findings
- This review: 3 findings
- Net change: -2 findings

📈 **Good progress!** You resolved 5 findings from the first review. One new finding was introduced in a fix commit, plus two gaps uncovered by the new validation code.

---

System of Review completed analysis of 'Retained-mode tilemap_scene Scene + Renderer with delta cache and sprite batching'. Found **3 findings** across 2 files.

**Author Feedback Addressed**: Your comment describing the 6 fix commits has been reviewed. Response included at the end.

## 📋 Review Summary

| 📖 DOCUMENTATION | 1 | 0 | 1 |
| ✅ TESTING | 0 | 1 | 1 |
| 🔍 QUALITY | 0 | 1 | 1 |
| **Total** | **1** | **2** | **3** |

---

## 📖 DOCUMENTATION

**DOCUMENTATION FINDING**: 🔴 Blocking | INTRODUCED — `pipeline.rs:113-115` `from_hex_size` doc (added by fix commit `267f6713`) claims Square4/Square8 are "not yet implemented — `load` will reject them at validation time." This is inaccurate: `ValidationError::UnsupportedTiling` is declared in `error.rs` but is never constructed anywhere in the codebase; `validate.rs:169-170` has the tiling check as `// TODO SPEC §16` (unimplemented); `load()` today returns `Ok(())` for Square specs. Rejection occurs in `gather_frame_emits` as `CompileError::UnsupportedAnchor` at render/compile time.

**Evidence:**
```rust
// pipeline.rs:113-115 (introduced by 267f6713):
/// `Square4` / `Square8` are accepted without panic (stride
/// defaults to `(w, h)`) but are not yet implemented —
/// [`crate::load::load`] will reject them at validation time.

// validate.rs:169-170 (unimplemented TODO):
// TODO SPEC §16: pipeline.tiling is one of the supported values
//                (HexFlatTop / HexPointyTop in 0.2.0; Square4 / Square8 rejected).

// ValidationError::UnsupportedTiling: declared in error.rs, never constructed
```

**Action:**
- [ ] Replace the false second sentence in the `from_hex_size` doc note with accurate text: tiling validation is a tracked TODO in `validate.rs` (SPEC §16); `load()` currently returns `Ok(())` for Square specs; compilation will fail with `CompileError::UnsupportedAnchor` at `render()` time. While in the file, also correct the analogous false claim in the `TilingStrategy` enum doc at `pipeline.rs:136-138` ("rejected at load time with `ValidationError::UnsupportedTiling`").


---

## ✅ TESTING

**TESTING FINDING**: 🟡 Non-blocking | INTRODUCED — Commit `3bbd7c48` adds `MissingDefaultState` and `ReservedId` enforcement to `validate.rs` (lines 148-162) but adds no negative-path tests for either rule in `tests/scene_model_test.rs`, despite the existing pattern of one dedicated negative-path test per enforced validation rule.

**Evidence:**
```
validate.rs:148-162 — two new enforcement checks added.

tests/scene_model_test.rs — existing negative-path tests:
  validate_rejects_unknown_pipeline_layer
  validate_rejects_unknown_asset_in_sprite_source
  validate_rejects_unknown_asset_in_animation

No corresponding tests for MissingDefaultState or ReservedId.
```

**Action:**
- [ ] Add `validate_rejects_missing_default_state` and `validate_rejects_reserved_id` negative-path tests to `tests/scene_model_test.rs`, following the existing pattern (construct a minimal `RenderSpec` that triggers exactly one violation, call `.validate()`, assert the expected `ValidationError` variant is returned).


---

## 🔍 QUALITY

**QUALITY FINDING**: 🟡 Non-blocking | INTRODUCED — `validate.rs:158` checks the reserved ID with inline string literal `"void"` while the codebase already exports `VOID_ID = "void"` from `compile::neighbors` and uses it for the same semantic purpose throughout `compile/`.

**Evidence:**
```rust
// validate.rs:158
if object.id == "void"

// compile/neighbors.rs:17
pub const VOID_ID : &str = "void";

// compile/neighbors.rs:247
exposed use VOID_ID;   // accessible as crate::compile::neighbors::VOID_ID from validate.rs
```

**Action:**
- [ ] Replace the inline literal `"void"` at `validate.rs:158` with `crate::compile::neighbors::VOID_ID` (or add a `use` import) to maintain a single source of truth for this reserved identifier.


---

This is a great contribution. The retained-mode migration is well-architected — the delta cache keyed on `(revision, clock.bits(), camera_sig)`, the fail-fast `Catalog` builder, and the `#[non_exhaustive]` `SceneEvent` all demonstrate careful API design. The quick turnaround on the five original findings was appreciated. Once F-7 is addressed the PR is ready to merge.

---

## 💬 Author Feedback Responses

### Response to Your Comment (EXPLANATION)

You described the 6 fix commits, specifically noting the MissingDefaultState check (`contains_key(&object.default_state)`, 7 lines) and the ReservedId check (`object.id == "void"`, 4 lines) in commit `3bbd7c48`.

Thank you for the clear accounting. All six commits were reviewed in detail. Five of the six are fully correct and complete — the implementations of MissingDefaultState and ReservedId are logically sound; the float semantics fix, the "consecutive" → "all" correction, and the label cleanup are all accurate. The one issue that emerged (`pipeline.rs:113-115`) was introduced by the fix commit for F-4 — it's a documentation-level error in the new note, not a logic error. Two non-blocking gaps (`TESTING` and `QUALITY`) were also uncovered by the new validation code itself.

---

FairlyBet:
@wanguardd 

# Review response

## 📖 DOCUMENTATION

### Finding: false claim that `load` rejects Square tilings

**Status:** ✅ Fixed in commit `cdbb59a0`.

**Verification of the finding:** Confirmed accurate.

- `ValidationError::UnsupportedTiling` is declared in `error.rs:115` and `Display`-printed in `error.rs:157`, but a code-wide grep shows zero construction sites.
- `validate.rs:169-170` keeps the tiling check as a `// TODO SPEC §16` and `RenderSpec::validate` never inspects `pipeline.tiling`.
- `load()` therefore returns `Ok( () )` for `Square4` / `Square8` specs.
- The actual failure point is `compile/frame.rs:157` inside `hex_world_pixel`, which yields `CompileError::UnsupportedAnchor { anchor: "Square (tiling strategy not implemented)" }` at render time.

**Fix applied (commit `cdbb59a0`):** Rewrote the misleading sentences in **three** locations — the two flagged by the review plus one analogous occurrence found while searching:

1. `pipeline.rs:113-115` — `HexConfig::from_hex_size` doc note. Now states that load-time rejection is a tracked TODO in `crate::validate` (SPEC §16), `load()` returns `Ok(())` for square specs today, and compilation fails at render time with `CompileError::UnsupportedAnchor`.
2. `pipeline.rs:136-138` — `TilingStrategy` enum doc. Same correction, and explicitly notes that `ValidationError::UnsupportedTiling` is declared for the future check but not yet constructed.
3. `error.rs:111-115` — `ValidationError::UnsupportedTiling` variant doc (not flagged in the review but carrying the identical false "rejected at load time" claim). Aligned with the other two.

`cargo check -p tilemap_scene` passes after the edits.

## ✅ TESTING

### Finding: missing negative-path tests for `MissingDefaultState` and `ReservedId`

**Status:** ✅ Fixed in commit `b4805075`.

**Verification of the finding:** Confirmed accurate.

- `validate.rs:148-162` (added by `3bbd7c48`) pushes `ValidationError::MissingDefaultState` when `object.default_state` is not a key of `object.states`, and `ValidationError::ReservedId` when `object.id == "void"`.
- `tests/scene_model_test.rs` already follows the convention of one dedicated `validate_rejects_*` test per enforced rule (`validate_rejects_unknown_pipeline_layer`, `validate_rejects_unknown_asset_in_sprite_source`, `validate_rejects_unknown_asset_in_animation`) but had none for the two new rules.

**Fix applied (commit `b4805075`):** Added two tests after `validate_rejects_unknown_asset_in_animation`, each matching the existing template (minimal RON spec triggering exactly one violation, `expect_err`, `matches!` assertion on the expected variant with field-level guards):

- `validate_rejects_missing_default_state` — declares an Object with `default_state: "missing"` while `states` only contains `"default"`; asserts the `MissingDefaultState { object: "grass", state: "missing" }` variant is returned.
- `validate_rejects_reserved_id` — declares an Object whose `id` is `"void"` (SPEC §15.1 reserved id); asserts the `ReservedId { id: "void" }` variant is returned.

`cargo test -p tilemap_scene --test scene_model_test validate_rejects` passes all five tests (3 existing + 2 new).

## 🔍 QUALITY

### Finding: `validate.rs:158` uses inline `"void"` instead of `VOID_ID`

**Status:** ✅ Fixed in commit `b4a9bd90`.

**Verification of the finding:** Confirmed accurate.

- `compile/neighbors.rs:17` defines `pub const VOID_ID : &str = "void"` and re-exports it at line 247.
- `compile/conditions.rs:8`, `compile/edges.rs:13`, `compile/vertex.rs:21`, and `compile/neighbors.rs:158` all consume it through the constant, never through a literal.
- `validate.rs:158` (added by `3bbd7c48`) compared `object.id == "void"`, the only remaining inline duplicate of the canonical id in the crate.
- Dependency direction is clean: nothing under `compile/` imports `validate`, so adding `validate → compile::neighbors` introduces no cycle.

**Fix applied (commit `b4a9bd90`):** Added `use crate::compile::neighbors::VOID_ID;` to the `validate.rs` module imports and changed the comparison to `object.id == VOID_ID`. Behaviour is identical; the change is purely a single-source-of-truth refactor.

`cargo test -p tilemap_scene --test scene_model_test validate_rejects` still passes all 5 tests.

wanguardd (changes requested):
## 📊 Review Summary

**Review #3**

**Progress since Review #2**:
- Previous review: 3 findings (1 blocking, 2 non-blocking)
- This review: 1 finding (0 blocking, 1 non-blocking)
- Net change: -2 findings

📈 **Good progress!** All three findings from the second review were addressed. One location was missed during the documentation fix.

---

Found **1 finding** across 1 file.

**Author Feedback Addressed**: Your comment describing the three fix commits has been reviewed. Response included at the end.

## 📋 Review Summary

| 📖 DOCUMENTATION | 0 | 1 | 1 |
| **Total** | **0** | **1** | **1** |

---

## 📖 DOCUMENTATION

**🟡 Non-blocking INTRODUCED** — `pipeline.rs:128` inline comment was not updated by `cdbb59a0`. The commit corrected the false "rejected by validation" claim in three documentation locations, but a fourth instance of the same claim remains in the function body:

**Evidence:**
```rust
// pipeline.rs:128-130 (NOT updated by cdbb59a0):
// Square tilings are rejected by validation; we still produce a
// sane default (1:1 with the bounding box) rather than panic so
// callers exploring the API don't trip a destructive failure.

// pipeline.rs:113-119 (correctly updated by cdbb59a0 — contradicts the above):
/// `Square4` / `Square8` are accepted without panic (stride
/// defaults to `(w, h)`) but are not yet implemented. Load-time
/// rejection of unsupported tilings is a tracked TODO in
/// [`crate::validate`] (SPEC §16), so [`crate::load::load`]
/// currently returns `Ok( () )` for square specs; compilation
/// later fails at render time with
/// [`crate::compile::CompileError::UnsupportedAnchor`].
```

The outer `///` doc (lines 113-119) is now accurate. The inline `//` comment one line below the match arm contradicts it — a contributor debugging why `load()` accepts square specs would find conflicting information within the same function.

**Action:**
- [ ] Update the inline comment at `pipeline.rs:128` to remove the false "rejected by validation" clause. The second and third lines of the comment (the rationale for producing a sane default rather than panicking) are correct and should be preserved.

---

## 💬 Author Feedback Responses

### Response to Your Comment (EXPLANATION)

You described the three fix commits, confirming the finding for each and detailing the specific changes made.

Thank you for the clear and detailed accounting. All three commits are correct:

- `cdbb59a0` corrected the documentation in `pipeline.rs:113-115`, `pipeline.rs:136-138`, and `error.rs:111-115` — all three updated locations are accurate and consistent with each other and with the actual implementation.
- `b4805075` adds `validate_rejects_missing_default_state` and `validate_rejects_reserved_id` — both follow the existing pattern, trigger exactly one violation, and make field-level assertions.
- `b4a9bd90` correctly routes the reserved-id comparison through `VOID_ID` with a clean import. The dependency direction is sound and the behaviour is identical.

The one remaining item is the inline `//` comment at `pipeline.rs:128`, which was not part of the three locations updated by `cdbb59a0`. It carries the same false claim as the now-corrected `///` doc directly above it.

FairlyBet:
@wanguardd 

# Review response

## 🔍 QUALITY

### Finding: stale "rejected by validation" comment left in `from_hex_size` body

**Status:** ✅ Fixed in commit `c34939c5`.

**Verification of the finding:** Confirmed accurate.

- `pipeline.rs:128` carried `// Square tilings are rejected by validation; …` — the exact false statement that commit `cdbb59a0` corrected at three documentation sites in the same file.
- The outer `///` doc on `from_hex_size` (lines 113-119, already fixed) and this inline `//` comment (still wrong) were contradicting each other inside the same function.
- The remaining two lines of the inline comment (rationale for emitting a 1:1 default rather than panicking) were and remain accurate.

**Fix applied (commit `c34939c5`):** Replaced only the first line of the inline comment — `"Square tilings are rejected by validation;"` → `"Square tilings are not yet implemented;"`. The rationale lines below are kept verbatim.

`cargo check -p tilemap_scene` passes after the edit.

wanguardd (changes requested):
## 📊 Review Summary

**Review #4**

**Progress since Review #3**:
- Previous review: 1 finding (0 blocking, 1 non-blocking)
- This review: 1 finding (0 blocking, 1 non-blocking)
- Net change: 0

The fix for the previous finding is correct and complete for the source file. One instance of the same false phrase survives in the companion test comment.

---

Found **1 finding** across 1 file.

**Author Feedback Addressed**: Your response to the previous review has been reviewed. Response included at the end.

## 📋 Review Summary

| 📖 DOCUMENTATION | 0 | 1 | 1 |
| **Total** | **0** | **1** | **1** |

---

## 📖 DOCUMENTATION

**🟡 Non-blocking INTRODUCED** — `tests/hex_config_test.rs:35` carries the same false "rejected by validation" claim that `c34939c5` corrected in `src/pipeline.rs:128`. The commit was scoped to exactly one file (`1 file changed, 1 insertion(+), 1 deletion(-)`), and the companion test comment was not updated.

**Evidence:**
```rust
// tests/hex_config_test.rs:35-37 (NOT updated by c34939c5):
// Square tilings are rejected by validation, but the constructor
// returns a sane (w, h) default so exploratory callers don't trip
// a panic.
```

The assertion on line 39 (`assert_eq!( cfg.grid_stride, ( 32, 32 ) )`) is correct — the constructor does return the bounding-box default. Only the comment is false: Square tilings are not rejected at validation time (`validate.rs:169-171` keeps this as a `// TODO SPEC §16`), `load()` returns `Ok(())` for them, and failure surfaces at render time with `CompileError::UnsupportedAnchor`.

**Action:**
- [ ] Update the comment at `tests/hex_config_test.rs:35` to remove the false "rejected by validation" clause. The rationale for the sane default is accurate and should be kept. Example first line: `// Square tilings are not yet rejected at load time (validation is a TODO);`

---

## 💬 Author Feedback Responses

### Response to Your Comment (EXPLANATION)

You verified the R2-1 finding and described the fix in `c34939c5` accurately: the first line of the three-line inline `//` comment was replaced (`"rejected by validation"` → `"not yet implemented"`), the two rationale lines were kept verbatim, and `cargo check -p tilemap_scene` passes.

All of that is confirmed correct. The one remaining item is the exact same phrase in the test comment at `tests/hex_config_test.rs:35`, which covers the same function (`from_hex_size`) with the same incorrect claim. The fix for the source and the fix for the test can be combined into one commit.

FairlyBet:
@wanguardd 

## DOCUMENTATION — tests/hex_config_test.rs:35 stale "rejected by validation" comment

**Diagnosis:** Correct. The comment introduced with the test still claimed Square tilings are
rejected by validation, but validate.rs:170 keeps that rule as a `// TODO SPEC §16` stub —
`load()` returns `Ok(())` for Square tilings and failure only surfaces at render time as
`CompileError::UnsupportedAnchor`.

**Fix applied:** Updated the three-line comment to read:

```rust
// Square tilings are not yet rejected at load time (validation is a TODO);
// the constructor still returns a sane (w, h) default so exploratory callers
// don't trip a panic.
```

**Commit:** `0b5e5643` — `docs(hex_config_test): correct Square tiling comment — not rejected at load time`

wanguardd (changes requested):
## Author Response — `0b5e5643`

The fix to `tests/hex_config_test.rs:35-37` is accurate. The updated comment ("Square tilings are not yet rejected at load time (validation is a TODO)") correctly describes the actual behavior: `load()` returns `Ok(())` for Square specs — `validate.rs:169-171` keeps the tiling-whitelist check as `// TODO SPEC §16` — and the failure surfaces at render time as `CompileError::UnsupportedAnchor`. The new wording is consistent with all corrected documentation in `pipeline.rs`.

---

## 📖 DOCUMENTATION

**🟡 Non-blocking** — `roadmap.md:182` carries the same false claim that has been corrected in `pipeline.rs:113-119`, `pipeline.rs:128`, `error.rs:111-118`, and `tests/hex_config_test.rs:35-37` across prior rounds:

\`\`\`
10. **Square tilings (\`Square4\` / \`Square8\`).** Enum values exist, rejected
    at load.
\`\`\`

"Rejected at load" is inaccurate. \`validate.rs:169-171\` keeps the tiling-whitelist rule as an unimplemented \`// TODO SPEC §16\` stub; \`load()\` returns \`Ok(())\` for \`Square4\` / \`Square8\` specs today, and failure only surfaces at render time as \`CompileError::UnsupportedAnchor\`. Compare item 3 in the same section ("A future minor version can implement them without breaking the format") — this item's treatment should reflect the same \`TODO\` status.

Suggested correction:

\`\`\`
10. **Square tilings (\`Square4\` / \`Square8\`).** Enum values exist; load-time
    rejection is a tracked TODO (SPEC §16) — \`load()\` currently returns
    \`Ok(())\`, and Square specs fail at render time with
    \`CompileError::UnsupportedAnchor\`. Implementing means square-grid neighbour
    offsets (4 or 8), square-grid dual-mesh (4 corners per vertex), square pixel
    conversion. Scope inflation — only do if a square-grid game is actually planned.
\`\`\`

---

## Summary

| 📖 DOCUMENTATION | 0 | 1 | 1 |
| **Total** | **0** | **1** | **1** |

FairlyBet:
@wanguardd 

# Review Response

## 🟡 Non-blocking — `roadmap.md:182` Square tiling "rejected at load" claim

**Finding: Correct.** The claim was inaccurate.

`validate.rs:170-171` confirms the tiling-whitelist rule is an unimplemented TODO stub:
```rust
// TODO SPEC §16: pipeline.tiling is one of the supported values
//                (HexFlatTop / HexPointyTop in 0.2.0; Square4 / Square8 rejected).
```
No enforcement code exists — `load()` returns `Ok(())` for Square4/Square8 specs.
Failure only surfaces at render time as `CompileError::UnsupportedAnchor`.

**Fix applied** in commit `684192dd`:
Updated item 10 in `roadmap.md` to reflect the actual TODO status, matching
the treatment of other unimplemented SPEC §16 rules.

wanguardd (approved):
## Author Response — `684192dd`

The fix to `roadmap.md:182-185` is accurate. The updated text — "Enum values exist; load-time rejection is a tracked TODO (SPEC §16) — `load()` currently returns `Ok(())`, and Square specs fail at render time with `CompileError::UnsupportedAnchor`" — correctly describes the implementation: `validate.rs:170-171` keeps the tiling-whitelist rule as an unimplemented `// TODO SPEC §16` stub, so `load()` returns `Ok(())` for Square specs today. The wording is consistent with `pipeline.rs:113-119` and `pipeline.rs:128`, and with the corrected `tests/hex_config_test.rs:35`.

---

## Summary

No new findings. All 12 findings from prior rounds are resolved:

| Finding | Resolved by |
|---------|------------|
| R0-1 `event.rs` stale "NOT re-arm" doc | `982d079b` |
| R0-2 `renderer.rs` CameraSignature NaN claim | `cc75d9da` |
| R0-3 `renderer.rs` "consecutive" merge | `a114503a` |
| R0-4 `pipeline.rs` Square accepted silently | `267f6713` + `cdbb59a0` |
| R0-5 `renderer.rs` opaque step labels | `f9399837` |
| R0-6 `validate.rs` MissingDefaultState/ReservedId unimplemented | `3bbd7c48` |
| F-7 `pipeline.rs:113` false "rejected at validation time" | `cdbb59a0` |
| N-1 `validate.rs` inline `"void"` literal | `b4a9bd90` |
| N-2 no negative-path tests for MissingDefaultState/ReservedId | `b4805075` |
| R2-1 `pipeline.rs:128` "rejected by validation" comment | `c34939c5` |
| R3-1 `hex_config_test.rs:35` "rejected by validation" comment | `0b5e5643` |
| R5-1 `roadmap.md:182` "rejected at load" | `684192dd` |

The codebase is internally consistent: Square specs pass `load()`, every documentation site now says so, and `CompileError::UnsupportedAnchor` at render time is the correctly-documented failure mode throughout. **Approved.**

PR #197, opened 2026-06-11
APPROVED : file loader: error propagation, blob/data URL support, inline buffer fix

FairlyBet:
## Problem

`file::load` and the glTF loader used `.expect()` / `.unwrap()` at every failure
point. In a WASM module, a panic aborts the entire module and leaves it permanently
broken — the user has to reload the page. Any transient error (network blip, typo in
a path, bad glTF file during development) triggered this abort.

Additionally, passing `blob:` or `data:` URLs to `load` was broken in two independent
ways that were not obvious from reading the code.

### `mingl` — `web/file.rs`

**Error propagation** (`load`):

- `Request::new_with_str_and_init` — was `.expect("Invalid url")`, now `?`
- `fetch_with_request` — was `.expect("Fetch request fail")`, now `?`
- `resp_value.dyn_into::<Response>()` — was `.unwrap()`, now `?`
- `window.location().origin()` — was `.unwrap()`, now `?`; the `# Panics` doc
  previously listed this as a panic site but the updated doc omitted it — fixed.

**`blob:` and `data:` URL support** (`resolve_url`):

- Both schemes are now passed through unchanged, just like `http://`/`https://`/`//`.
  Without this, `blob:https://...` or `data:...` would have been treated as
  origin-relative paths and prefixed with the window origin, producing an
  unresolvable URL.
- Two new unit tests cover the passthrough behaviour.

**`data:` URL decoding** (`load`):

- Browsers reject `fetch()` with `RequestMode::Cors` when the URL is a `data:` URL
  (the Fetch spec disallows CORS mode for non-HTTP schemes; Chrome and Firefox both
  throw `TypeError: Failed to fetch`). Simply adding `data:` to the passthrough list
  in `resolve_url` was therefore not enough — the fetch would still fail.
- `data:` URLs are now short-circuited before the `RequestInit` is constructed:
  the base64 payload is decoded via `window.atob()` and returned directly. This
  avoids the network entirely, which is also faster.
- Only `;base64`-encoded data URLs are supported; non-base64 payloads return `Err`.

**Cleanup**:

- `vec![0; len]; uint8_array.copy_to(&mut data[..])` → `uint8_array.to_vec()`.
  The old form zero-initialised the buffer and then overwrote every byte; `to_vec`
  allocates uninitialised capacity and writes once.

### `renderer` — `webgl/loaders/gltf.rs`

**Error propagation** (glTF / buffer loading):

- Loading the root glTF file — was `.expect(...)` + `.unwrap()`, now
  `.map_err(|_| WebglError::Other(...))?` for both the fetch and the parse step.
- Loading each external buffer — was `.expect(...)`, now
  `.map_err(|_| WebglError::Other(...))?`.

**Inline buffer (`data:` URI) fix**:

- The glTF spec allows `buffer.uri` to be a `data:` URI (a base64-encoded inline
  buffer). The loader unconditionally did `format!("{}/{}", folder_path, uri)`,
  which prepended the folder path to the raw URI string — producing something like
  `"models/data:application/octet-stream;base64,..."`. This garbled string reached
  `resolve_url` as an origin-relative path and the fetch failed.
- The buffer path construction now checks the URI scheme first and only prepends
  `folder_path` for relative paths; absolute URLs, `blob:`, `data:`, and
  origin-absolute (`/`) URIs are used verbatim.

## Behaviour that did not change

- HTTP 4xx/5xx responses still resolve as `Ok` (the body of the error page is
  returned). This is intentional and documented: `fetch` resolves successfully for
  any HTTP response; callers that care about the status must inspect it themselves.
- `web_sys::window().unwrap()` still panics when called outside a browsing context
  (Worker, SSR). This is documented and intentional — the function is explicitly
  a browser-only API.

### DIFF
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -421,8 +421,14 @@ mod private
 
     // let gltf_slice= gl::file::load( &format!( "{}/scene.gltf", gltf_path ) )
     // .await.expect( "Failed to load gltf file" );
-    let gltf_slice = gl::file::load( gltf_path ).await.expect( "Failed to load gltf file" );
-    let mut gltf_file = gltf::Gltf::from_slice( &gltf_slice ).unwrap();
+    // Propagate fetch / parse failures as errors instead of panicking: an
+    // `.unwrap()` here aborts the whole wasm module (e.g. when a dev server
+    // returns an HTML 404 page, or the bytes are not a valid glTF/GLB), leaving
+    // it unusable for every subsequent call.
+    let gltf_slice = gl::file::load( gltf_path ).await
+    .map_err( | _ | gl::WebglError::Other( "Failed to load gltf file" ) )?;
+    let mut gltf_file = gltf::Gltf::from_slice( &gltf_slice )
+    .map_err( | _ | gl::WebglError::Other( "Failed to parse gltf file" ) )?;
 
     let mut buffers : Vec< gl::js_sys::Uint8Array > = Vec::new();
 
@@ -442,7 +448,7 @@ mod private
         {
           let path = format!( "{}/{}", folder_path, uri );
           let buffer = gl::file::load( &path ).await
-          .expect( "Failed to load a buffer" );
+          .map_err( | _ | gl::WebglError::Other( "Failed to load a buffer" ) )?;
 
           gl::debug!
           (
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -436,6 +436,75 @@ mod private
     }
   }
 
+  #[ cfg( test ) ]
+  mod tests
+  {
+    use super::resolve_asset_uri;
+
+    #[ test ]
+    fn joins_relative_uri_with_folder()
+    {
+      assert_eq!
+      (
+        resolve_asset_uri( "models", "scene/buffer.bin" ),
+        "models/scene/buffer.bin"
+      );
+    }
+
+    #[ test ]
+    fn passes_blob_uri_through()
+    {
+      assert_eq!
+      (
+        resolve_asset_uri( "models", "blob:https://app.example.com/uuid-1234" ),
+        "blob:https://app.example.com/uuid-1234"
+      );
+    }
+
+    #[ test ]
+    fn passes_data_uri_through()
+    {
+      assert_eq!
+      (
+        resolve_asset_uri( "models", "data:application/octet-stream;base64,Z2xURg==" ),
+        "data:application/octet-stream;base64,Z2xURg=="
+      );
+    }
+
+    #[ test ]
+    fn passes_absolute_url_through()
+    {
+      assert_eq!
+      (
+        resolve_asset_uri( "models", "https://cdn.example.com/textures/t.png" ),
+        "https://cdn.example.com/textures/t.png"
+      );
+    }
+
+    #[ test ]
+    fn passes_origin_absolute_path_through()
+    {
+      assert_eq!
+      (
+        resolve_asset_uri( "models", "/textures/t.png" ),
+        "/textures/t.png"
+      );
+    }
+
+    #[ test ]
+    fn empty_folder_yields_origin_absolute_uri()
+    {
+      // Documents the benign empty-folder behavior: origin-absolute and
+      // origin-relative forms collapse to the same URL once `resolve_url`
+      // joins them against the window origin.
+      assert_eq!
+      (
+        resolve_asset_uri( "", "buffer.bin" ),
+        "/buffer.bin"
+      );
+    }
+  }
+
   /// Asynchronously loads a glTF (GL Transmission Format) file and its associated resources.
   pub async fn load
   (
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -577,9 +577,28 @@ mod private
         }
       );
 
+      // Without an onerror handler a 404 or malformed image URI fails silently:
+      // the 1x1 white placeholder stays bound, nothing is logged, and load()
+      // still returns Ok. Mirror the error logging added for buffer URI loads so
+      // image failures are diagnosable instead of rendering as blank textures.
+      let on_error : Closure< dyn Fn() > = Closure::new
+      (
+        {
+          let img = img_element.clone();
+          let src = src.clone();
+          move ||
+          {
+            gl::browser::error!( "Failed to load gltf image '{src}'" );
+            img.remove();
+          }
+        }
+      );
+
       img_element.set_onload( Some( load_texture.as_ref().unchecked_ref() ) );
+      img_element.set_onerror( Some( on_error.as_ref().unchecked_ref() ) );
       img_element.set_src( &src );
       load_texture.forget();
+      on_error.forget();
     };
 
     // If a source of an image is Uri - load the file
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -405,6 +405,32 @@ mod private
     )
   }
 
+  /// Resolves a glTF asset `uri` (buffer or image) against the model's `folder_path`.
+  ///
+  /// URIs that already carry their own location are returned unchanged, because
+  /// prefixing `folder_path` would corrupt them:
+  /// * absolute / protocol-relative URLs (`http://`, `https://`, `//`),
+  /// * self-contained URIs (`blob:`, `data:`),
+  /// * origin-absolute paths (leading `/`).
+  ///
+  /// Everything else is treated as folder-relative and joined with a single `/`.
+  fn resolve_asset_uri( folder_path : &str, uri : &str ) -> String
+  {
+    if uri.starts_with( "http://" )
+    || uri.starts_with( "https://" )
+    || uri.starts_with( "//" )
+    || uri.starts_with( "blob:" )
+    || uri.starts_with( "data:" )
+    || uri.starts_with( '/' )
+    {
+      uri.to_string()
+    }
+    else
+    {
+      format!( "{}/{}", folder_path, uri )
+    }
+  }
+
   /// Asynchronously loads a glTF (GL Transmission Format) file and its associated resources.
   pub async fn load
   (
@@ -446,21 +472,7 @@ mod private
       {
         gltf::buffer::Source::Uri( uri ) =>
         {
-          // Absolute URLs, data: and blob: URIs, and origin-absolute paths carry their
-          // own location — prepending folder_path would destroy the scheme or leading slash.
-          let path = if uri.starts_with( "http://" )
-          || uri.starts_with( "https://" )
-          || uri.starts_with( "//" )
-          || uri.starts_with( "blob:" )
-          || uri.starts_with( "data:" )
-          || uri.starts_with( '/' )
-          {
-            uri.to_string()
-          }
-          else
-          {
-            format!( "{}/{}", folder_path, uri )
-          };
+          let path = resolve_asset_uri( folder_path, uri );
           let buffer = gl::file::load( &path ).await
           .map_err( | _ | gl::WebglError::Other( "Failed to load a buffer" ) )?;
 
@@ -559,7 +571,7 @@ mod private
       {
         gltf::image::Source::Uri { uri, mime_type: _ } =>
         {
-          upload_texture( format!( "{}/{}", folder_path, uri ).into() );
+          upload_texture( resolve_asset_uri( folder_path, uri ).into() );
         },
         gltf::image::Source::View { view, mime_type } =>
         {
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -416,12 +416,10 @@ mod private
   /// Everything else is treated as folder-relative and joined with a single `/`.
   fn resolve_asset_uri( folder_path : &str, uri : &str ) -> String
   {
-    if uri.starts_with( "http://" )
-    || uri.starts_with( "https://" )
-    || uri.starts_with( "//" )
-    || uri.starts_with( "blob:" )
-    || uri.starts_with( "data:" )
-    || uri.starts_with( '/' )
+    // `gl::file::load` already resolves self-contained URLs and origin-absolute
+    // paths against the window origin; only genuinely folder-relative URIs need
+    // the model's folder prefix folded in.
+    if gl::file::is_self_contained_url( uri ) || uri.starts_with( '/' )
     {
       uri.to_string()
     }
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -449,10 +449,22 @@ mod private
     // `.unwrap()` here aborts the whole wasm module (e.g. when a dev server
     // returns an HTML 404 page, or the bytes are not a valid glTF/GLB), leaving
     // it unusable for every subsequent call.
+    // `WebglError::Other` only carries a `&'static str`, so the underlying
+    // `JsValue` / `gltf::Error` (file path, HTTP status, JSON parse location)
+    // would otherwise be lost. Log it to the console before mapping so a failed
+    // load is diagnosable in production.
     let gltf_slice = gl::file::load( gltf_path ).await
-    .map_err( | _ | gl::WebglError::Other( "Failed to load gltf file" ) )?;
+    .map_err( | e |
+    {
+      gl::browser::error!( "Failed to load gltf file '{gltf_path}': {e:?}" );
+      gl::WebglError::Other( "Failed to load gltf file" )
+    } )?;
     let mut gltf_file = gltf::Gltf::from_slice( &gltf_slice )
-    .map_err( | _ | gl::WebglError::Other( "Failed to parse gltf file" ) )?;
+    .map_err( | e |
+    {
+      gl::browser::error!( "Failed to parse gltf file '{gltf_path}': {e}" );
+      gl::WebglError::Other( "Failed to parse gltf file" )
+    } )?;
 
     let mut buffers : Vec< gl::js_sys::Uint8Array > = Vec::new();
 
@@ -472,7 +484,11 @@ mod private
         {
           let path = resolve_asset_uri( folder_path, uri );
           let buffer = gl::file::load( &path ).await
-          .map_err( | _ | gl::WebglError::Other( "Failed to load a buffer" ) )?;
+          .map_err( | e |
+          {
+            gl::browser::error!( "Failed to load gltf buffer '{path}': {e:?}" );
+            gl::WebglError::Other( "Failed to load a buffer" )
+          } )?;
 
           gl::debug!
           (
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -564,7 +564,13 @@ mod private
             gl.generate_mipmap( gl::TEXTURE_2D );
             gl.tex_parameteri( gl::TEXTURE_2D, gl::TEXTURE_MIN_FILTER, gl::LINEAR_MIPMAP_LINEAR as i32 );
 
-            gl::web_sys::Url::revoke_object_url( &src ).unwrap();
+            // revoke_object_url is specified only for blob: URLs; for data: URIs or
+            // plain file paths it is a no-op, and unwrapping its result is a latent
+            // panic hazard in stricter runtimes. Only revoke the urls we created.
+            if src.starts_with( "blob:" )
+            {
+              gl::web_sys::Url::revoke_object_url( &src ).unwrap();
+            }
 
             img.remove();
           }
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -414,6 +414,13 @@ mod private
   /// * origin-absolute paths (leading `/`).
   ///
   /// Everything else is treated as folder-relative and joined with a single `/`.
+  ///
+  /// When `folder_path` is empty (the glTF was loaded from a bare filename, so it
+  /// sits at the origin root) a folder-relative `uri` resolves to `"/{uri}"`. This
+  /// is intentional and harmless: `resolve_url` joins both `"/buffer.bin"` and
+  /// `"buffer.bin"` against the origin to the same `"{origin}/buffer.bin"`. A glTF
+  /// served from a subdirectory must be loaded with that directory in `gltf_path`
+  /// (e.g. `"assets/scene.gltf"`), otherwise the glTF fetch itself fails first.
   fn resolve_asset_uri( folder_path : &str, uri : &str ) -> String
   {
     // `gl::file::load` already resolves self-contained URLs and origin-absolute
--- a/module/helper/renderer/src/webgl/loaders/gltf.rs
+++ b/module/helper/renderer/src/webgl/loaders/gltf.rs
@@ -446,7 +446,21 @@ mod private
       {
         gltf::buffer::Source::Uri( uri ) =>
         {
-          let path = format!( "{}/{}", folder_path, uri );
+          // Absolute URLs, data: and blob: URIs, and origin-absolute paths carry their
+          // own location — prepending folder_path would destroy the scheme or leading slash.
+          let path = if uri.starts_with( "http://" )
+          || uri.starts_with( "https://" )
+          || uri.starts_with( "//" )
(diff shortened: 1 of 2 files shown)


wanguardd (changes requested):
## 🐛 CORRECTNESS

🔴 Image URI loop not updated alongside buffer URI loop — INTRODUCED

**Evidence:**
```json
The diff fixes the buffer URI path construction in gltf.rs (lines 449–462) with a scheme-detection guard, but the image URI loop immediately below is unchanged: `upload_texture(format!("{}/{}", folder_path, uri).into())`. Any GLTF file with embedded (`data:`) or externally-hosted (`blob:`, `http://`, absolute `/`) image textures will receive a mangled path (e.g. `"models/data:image/png;base64,..."`) and silently fail. The texture upload closure initialises a 1×1 white placeholder, so the failure mode is silent: broken textures, no error propagated.
```

**Action:**
- [ ] Apply the same scheme-detection block to the image URI loop. The simplest fix is a shared helper `fn resolve_asset_uri(folder_path: &str, uri: &str) -> String` used by both loops — this also eliminates the duplication noted separately below.


---

## 📐 DESIGN

`map_err(|_| WebglError::Other(...))` discards original error context — INTRODUCED

**Evidence:**
```
Every changed error site in gltf.rs uses `map_err(|_| gl::WebglError::Other("..."))`, where `_` discards the underlying `gltf::Error` or JS `TypeError`/`DOMException`. `WebglError::Other` holds `&'static str`, so dynamic context (file path, HTTP status, JSON parse location) cannot be included. When a load fails in production the only observable information is a fixed English string with no location.
```

**Action:**
- [ ] Change the `Other` variant to carry a `String`: `Other(String)`, then use `.map_err(|e| gl::WebglError::Other(format!("...{e}")))` at each call site. Alternatively, log the original error via `web_sys::console::error_1(&e)` before discarding it with `map_err(|e| { web_sys::console::error_1(&e); ... })`.


---

## 🔍 QUALITY

URL scheme detection duplicated between file.rs and gltf.rs — INTRODUCED

**Evidence:**
```
`file.rs` lines 17–21 enumerate 5 pass-through patterns (`http://`, `https://`, `//`, `blob:`, `data:`). `gltf.rs` lines 449–456 enumerate 6 patterns (the same 5 plus `'/'` for origin-absolute paths). The two lists have already diverged (the `/` case). Any scheme added to one list in future must be manually added to the other.
```

**Action:**
- [ ] Extract a shared helper — e.g. `fn resolve_asset_uri(base: &str, uri: &str) -> String` — accessible from both crates, or at minimum within the renderer crate. Call it from both the buffer and image loops in gltf.rs, and review whether resolve_url in file.rs can be reused or delegated to the same helper.


---

## ✅ TESTING

No wasm_bindgen_test coverage for the data: decode path in load() — INTRODUCED

**Evidence:**
```
The `#[cfg(test)] mod tests` block in file.rs tests only `resolve_url` (a pure-Rust function). The `load()` function's new `data:` branch — comma detection, base64 check, `window.atob()` invocation, and the `chars().map(|c| c as u8)` byte extraction — has zero automated test coverage. Edge cases (malformed header, non-base64 encoding, missing comma, empty payload) are reachable in production but not exercised.
```

**Action:**
- [ ] Add `#[wasm_bindgen_test]` tests (in `tests/` per cgtools conventions) covering: (1) valid base64 data URL returns correct bytes, (2) data URL without comma returns `Err`, (3) non-base64 encoding returns `Err`. A runner configuration for `wasm-pack test --headless --firefox` or equivalent is needed; add to CI if not already present.


---

---

>
>

wanguardd:
No new commits detected since the previous review. If you would like a re-review without new commits, set the PR title prefix to NEED REVIEW and invoke /pr_review 197 again.

FairlyBet:
@wanguardd 

# AI Review Processing — gltf/file `data:` URL loader (branch: simple-filter)

Review targets commit `a88607d5` ("fix data: URL handling and error propagation in
file loader") and the surrounding branch diff vs `master`. Each comment treated as a
hypothesis and verified against the real code before acting. One commit per item.

Affected files:

- `module/helper/renderer/src/webgl/loaders/gltf.rs`
- `module/min/mingl/src/web/file.rs`

| # | Item (short) | Status | Commit | Notes |
|---|--------------|--------|--------|-------|
| 1 | Image URI loop not updated alongside buffer URI loop | **FIXED** | `fb91087a` | Real bug. Extracted `resolve_asset_uri`, used in both loops. |
| 2 | `map_err(\|_\| Other(...))` discards original error context | **FIXED (best fix, not suggested)** | `bc4292e2` | Valid. Log original error; rejected the `Other(String)` API change. |
| 3 | URL scheme detection duplicated between file.rs and gltf.rs | **FIXED (partial)** | `3c82cf61` | Scheme-list dup is real; extracted shared predicate. The `/` "divergence" is by design, not a bug. |
| 4 | No test coverage for the `data:` decode path in `load()` | **FIXED (scoped)** | `91d07993` | Valid. Extracted pure parser + 4 unit tests; rejected standing up a wasm-bindgen-test browser runner. |

## Verification

- `cargo test -p mingl --features "web,web_future,web_file" --lib` → **13 passed** (was 9; +4 new).
- `cargo check -p renderer --target wasm32-unknown-unknown` → clean after every item.

## Per-item detail

### 1 — Image URI loop not updated (FIXED, `fb91087a`)

**Real bug.** The buffer loop got the scheme-detection guard but the image loop kept
`upload_texture( format!( "{}/{}", folder_path, uri ).into() )`. A `data:`/absolute/
origin-absolute image URI was mangled (e.g. `"models/data:image/png;..."`) and silently
fell back to the 1×1 white placeholder.

**Fix (better than suggested):** extracted `resolve_asset_uri(folder_path, uri)` and call
it from **both** the buffer and image loops, so they cannot diverge again. This is the
shared-helper the reviewer suggested, scoped to the renderer crate.

### 2 — `map_err(|_| Other(...))` discards context (FIXED, `bc4292e2`)

**Valid concern.** Three sites (`load`, `from_slice`/parse, buffer `load`) discard the
underlying `JsValue` / `gltf::Error`; `Other(&'static str)` cannot carry dynamic context.

**Rejected the primary suggestion** (`Other(String)`): the variant is used at ~21 call
sites across published `min` crates, and **every** variant in `WebglError` intentionally
uses `&'static str` (house style). Widening it is a breaking API change disproportionate
to this PR.

**Applied the alternative:** log the original error (with the path) via
`gl::browser::error!` before mapping, at all three sites. Note: `gl::error!` does **not**
exist — the `error` macro name collides with the `error` *module* (the `error` layer),
which shadows the glob-imported macro; the house-style path is `gl::browser::error!`
(already used in `pmrem.rs`/`shadow.rs`). Typed error preserved for callers; full
diagnostic now reaches the console.

### 3 — Scheme detection duplicated between file.rs and gltf.rs (FIXED partial, `3c82cf61`)

**Partially valid.** The 5-scheme list (`http://`, `https://`, `//`, `blob:`, `data:`)
was genuinely duplicated between `resolve_url` (mingl) and `resolve_asset_uri` (gltf) and
could drift. Extracted a single `is_self_contained_url` predicate in `mingl::web::file`,
exposed via `mod_interface`, and called from both (`gl::file::is_self_contained_url`).

**Rejected the framing that the `/` case is a drift bug.** The leading-`/` handling
differs *by design*: `resolve_url` joins origin-absolute paths to the window origin,
whereas the gltf helper passes them through (so `file::load` resolves them later).
Folding `/` into the shared predicate would have been wrong — it stays local to each
function. Did not collapse `resolve_url` and `resolve_asset_uri` into one function: they
have distinct join rules (origin vs folder).

### 4 — No coverage for the `data:` decode path (FIXED scoped, `91d07993`)

**Valid.** The comma-detection and base64-check edge cases were exercised only in
production.

**Rejected the suggested wasm-bindgen-test + CI browser runner:** the repo has **no**
`wasm-bindgen-test` dependency or headless-browser runner anywhere in `module/min`, and
one cannot be validated in this environment. Adding unverified harness + CI config is out
of scope and risky.

**Applied the testable refactor (matches the existing `resolve_url` pattern):** extracted
the pure validation/splitting into `data_url_base64_payload(url) -> Result<&str,&'static
str>` and added 4 native `#[test]`s — valid payload, empty payload, missing comma,
non-base64. `load()` now calls the helper; only the `window.atob` byte decode (genuinely
browser-only) remains uncovered, which is stated honestly rather than faked.

## Similar-occurrence sweep

- `module/helper/renderer/src/webgl/animation/loaders/gltf.rs`: receives already-decoded
  `buffers: &[Vec<u8>]` and builds only animation-channel keys — **no** asset-URI
  concatenation, no analogous bug.
- All other `file::load` callers (`hdr_texture.rs`, examples) pass caller-supplied paths
  and now inherit the centralized scheme detection in `resolve_url` automatically.

wanguardd (changes requested):
## Review Summary

| Attribute | Value |
|-----------|-------|
| Round | #2 |
| Pipeline | 399792a |
| Rulebook | v3.20 |
| Timestamp | 2026-06-21T00:44:26Z |
| Order | 1967 |

**Progress since Review #1**:
- Previous review: 4 findings
- This review: 8 findings
- All 4 prior findings addressed ✓

📝 **Prior findings resolved**: You closed all 4 findings from the previous review — `resolve_asset_uri` extraction, error logging, scheme-detection sharing, and `data_url_base64_payload` extraction with tests. The direction is right.

---

System of Review completed analysis of 'NEED REVIEW : file loader: error propagation, blob/data URL support, inline buffer fix'. Found **8 findings** across 2 files.

**Author Feedback Addressed**: Your 1 comment has been reviewed. Response is included below.

## 📋 Review Summary

| 🐛 CORRECTNESS | 0 | 2 | 2 |
| ✅ TESTING | 1 | 0 | 1 |
| 🔍 QUALITY | 0 | 1 | 1 |
| 📖 DOCUMENTATION | 0 | 3 | 3 |
| 📐 DESIGN | 0 | 1 | 1 |
| **Total** | **1** | **7** | **8** |

---

## 🐛 CORRECTNESS

`revoke_object_url` called unconditionally on all image sources, not just blob URLs

**Evidence:**
```rust
// gltf.rs ~line 567, inside upload_texture onload callback:
gl::web_sys::Url::revoke_object_url( &src ).unwrap();
```
This runs for every image load — HTTP paths, data: URIs, and blob: URLs alike. The Web API silently ignores `revokeObjectURL` on non-blob URLs, so there is no current error. However the logic is semantically wrong: `revoke_object_url` exists to release blob: object URLs created by `create_object_url_with_blob`. Calling it on `"static/textures/albedo.png"` or on a `data:` URI is a no-op that happens to succeed. This PR adds `data:` as a new reachable image source type, widening the set of non-blob URLs taking this path.

**Action:**
- [ ] Track whether the URL was blob-created (e.g., a `bool is_blob_url` alongside `src`, or an `Option<String>` holding the blob URL only when a blob was created) and call `revoke_object_url` only in that branch.

---

Empty `folder_path` produces an origin-absolute path, silently wrong when glTF is not at the server root

**Evidence:**
```rust
// When gltf_path = "scene.gltf" (bare filename, no directory component):
// Path::new("scene.gltf").parent() -> Some("")
// folder_path = ""
// resolve_asset_uri("", "buffer.bin") -> format!("{}/{}", "", "buffer.bin") -> "/buffer.bin"
// resolve_url("/buffer.bin") -> "https://origin.example.com/buffer.bin"  ← WRONG if gltf lives at /assets/
```
If a caller loads `"scene.gltf"` but that file is actually served from `https://example.com/assets/scene.gltf`, the buffer resolves to `https://example.com/buffer.bin` (404) instead of `https://example.com/assets/buffer.bin`. The `map_err` logging now reports the failure but the message "Failed to load gltf buffer" gives no hint that the path is wrong.

**Action:**
- [ ] Document the constraint that `gltf_path` must include its directory component (e.g., `"assets/scene.gltf"` not `"scene.gltf"`) in the function's doc comment. Alternatively, detect the empty-string case in `resolve_asset_uri` and either propagate a clearer error or document the caller's responsibility.

---

## ✅ TESTING

🔴 `resolve_asset_uri` is untested — a pure function on the critical path for all buffer and image loads

**Evidence:**
```
resolve_asset_uri(folder_path: &str, uri: &str) -> String
```
This function is synchronous, has no browser dependencies, and is directly testable with `cargo test`. It sits on the critical path for every buffer load (line ~485) and every image load (line ~588). The image failure mode is particularly severe: a bug in `resolve_asset_uri` for image URIs produces blank white textures with no console output and no error returned from `load()` (no `onerror` handler is registered — see QUALITY finding below).

Cases that need pinning:
- Standard relative URI: `resolve_asset_uri("models", "scene/buffer.bin")` → `"models/scene/buffer.bin"`
- Self-contained passthrough: `resolve_asset_uri("models", "blob:...")` → `"blob:..."`
- Self-contained passthrough: `resolve_asset_uri("models", "data:...")` → `"data:..."`
- Origin-absolute passthrough: `resolve_asset_uri("models", "/textures/t.png")` → `"/textures/t.png"`
- Empty `folder_path`: `resolve_asset_uri("", "buffer.bin")` → `"/buffer.bin"` (documents current behavior, see CORRECTNESS finding)

**Action:**
- [ ] Add a `#[cfg(test)] mod tests` block in `gltf.rs` (consistent with placement in the same file where the function is defined) covering at minimum the five cases above. The function is private, so inline tests are appropriate per the workspace pattern.

---

## 🔍 QUALITY

Image URI load failures are completely silent — no `onerror` handler on the `` element

**Evidence:**
```
// gltf.rs upload_texture closure:
img_element.set_src(&src);   // async browser load — no onerror registered
// If src is a 404 or malformed URL, browser fires onerror — no handler catches it.
// The 1×1 white placeholder texture remains; no error is logged; load() returns Ok(()).
```
The PR correctly adds error logging for buffer URI failures (new `gl::browser::error!` calls before `map_err`). Image URI failures remain completely silent. A model with a bad buffer URI fails fast with a logged error; the same model with a bad image URI silently renders with white textures and no console output.

**Action:**
- [ ] Register an `onerror` callback on `img_element` that calls `gl::browser::error!` with the failed URL. This matches the error-logging pattern added for buffer loads in this PR and closes the asymmetry between buffer and image error observability.

---

## 📖 DOCUMENTATION

`is_self_contained_url` doc incorrectly states that `data:` URLs reach the network layer

**Evidence:**
```rust
/// Returns `true` for URLs that must reach the network layer verbatim.
```
The doc says these URLs "must reach the network layer verbatim." This holds for `http://`, `https://`, `//`, and `blob:` — all of which go through `fetch`. However, `data:` URLs are explicitly intercepted *before* `fetch` (the branch at line ~113 in `file.rs`) precisely because `fetch()` rejects CORS mode for `data:` URIs. A `data:` URL never reaches the network layer.

**Action:**
- [ ] Revise the doc to distinguish the two subcategories: "absolute URLs (`http://`, `https://`, `//`, `blob:`) that go through `fetch` unchanged, and self-contained payloads (`data:`) that are decoded inline without a network request."

---

`resolve_url` doc lists only `blob:`/`data:` as pass-through schemes, omitting `http://`, `https://`, and `//`

**Evidence:**
```rust
// doc comment on resolve_url:
/// Self-contained URLs (`blob:`, `data:`) pass through unchanged.
```
The implementation delegates to `is_self_contained_url` which also passes `http://`, `https://`, and `//` through unchanged. A reader following the doc would conclude that an HTTP URL gets prefix-joined with the origin, which is incorrect.

**Action:**
- [ ] Expand the bullet: "Absolute URLs (`http://`, `https://`, `//`) and self-contained URLs (`blob:`, `data:`) pass through unchanged."

---

`atob` → `chars()` → `as u8` decode is correct but relies on an undocumented Web IDL invariant

**Evidence:**
```rust
// file.rs:
let decoded = window.atob( &b64 ).map_err( |_| "atob failed" )?;
return Ok( decoded.chars().map( | c | c as u8 ).collect() );
```
`window.atob()` returns a DOMString whose code points are Latin-1 bytes (U+0000–U+00FF per Web IDL spec). wasm-bindgen transports this as UTF-8, where each U+0080–U+00FF code point becomes a 2-byte UTF-8 sequence. `.chars()` re-assembles Unicode scalars (always in U+0000–U+00FF for valid atob output), and `c as u8` truncates to the low 8 bits — which is correct precisely because these code points fit in a byte. The decode is correct, but a future maintainer seeing `chars().map(|c| c as u8)` without context could plausibly "simplify" it (e.g., to byte-length arithmetic) and silently break binary payloads.

**Action:**
- [ ] Add a comment above the `chars().map(|c| c as u8)` line explaining the invariant: "`atob` always returns Latin-1 code points (U+0000–U+00FF); each `char` scalar fits in a `u8`, so truncation is lossless." This matches the existing pattern of explanatory comments elsewhere in `file.rs`.

---

## 📐 DESIGN

File `module/helper/renderer/src/webgl/loaders/gltf.rs` introduces large changes requiring evaluation

**Evidence:**
```
File contains 1105 lines with substantial changes across the diff
```

**Action:**
- [ ] Verify the changes in `module/helper/renderer/src/webgl/loaders/gltf.rs` align with the project architecture and that the single-file scope is intentional (the buffer load path, image load path, and glTF parse path are all modified together).

---

This is good incremental work. The shared `is_self_contained_url` predicate is solid design, and `data_url_base64_payload` is well-factored and tested. The main items before merge are: add tests for `resolve_asset_uri` (a pure function that currently has no test coverage despite sitting on a silent-failure image path) and fix the `revoke_object_url` logic. The three documentation findings can be done in a single small commit. Looking forward to the next iteration.

---

### Response to Your Comment #1 (EXPLANATION)

> 
@wanguardd 

# AI Review Processing — gltf/file `data:` URL loader (branch: simple-filter)

Review targets commit `a88607d5` ("fix data: URL handling and error propagation in
file loader") and the surrounding branch diff vs `master`. Each comment treated as a
hypothesis and verified against the real code before acting. One commit per item.

Affected files:

- `module/helper/renderer/src/webgl/loaders/gltf.rs`
- `module/min/mingl/src/web/file.rs`

| # | Item (short) | Status | Commit | Notes |
|---|--------------|--------|--------|-------|
| 1 | Image URI loop not updated alongside buffer URI loop | **FIXED** | `fb91087a` | Real bug. Extracted `resolve_asset_uri`, used in both loops. |
| 2 | `map_err(\|_\| Other(...))` discards original error context | **FIXED (best fix, not suggested)** | `bc4292e2` | Valid. Log original error; rejected the `Other(String)` API change. |
| 3 | URL scheme detection duplicated between file.rs and gltf.rs | **FIXED (partial)** | `3c82cf61` | Scheme-list dup is real; extracted shared predicate. The `/` "divergence" is by design, not a bug. |
| 4 | No test coverage for the `data:` decode path in `load()` | **FIXED (scoped)** | `91d07993` | Valid. Extracted pure parser + 4 unit tests; rejected standing up a wasm-bindgen-test browser runner. |

Acknowledged in review.

---

wanguardd (approved):
## 📐 DESIGN

📐 File module/helper/renderer/src/webgl/loaders/gltf.rs introduces large changes requiring evaluation.

**Evidence:**
```
File contains 1105 lines, 34313 bytes of changes
```

**Action:**
- [ ] Review the changes in module/helper/renderer/src/webgl/loaders/gltf.rs to ensure they align with the project architecture and consider splitting into smaller, focused commits if the changes span multiple concerns.


---

## 🐛 CORRECTNESS

`Url::revoke_object_url(&src).unwrap()` is called unconditionally inside the `onload` closure of `upload_texture` (gltf.rs ~line 518). This function is specified only for `blob:` URLs; per the W3C Blob URL spec, calling it with any other URL scheme is a no-op in Chrome and Firefox today — but calling `.unwrap()` on the result is a latent panic hazard in stricter runtimes (JSDOM, embedded WebView engines, future browser versions).

Before this PR, `data:` image URIs were always mangled by the folder-prefix and never reached `upload_texture` in a usable form. After this PR, `resolve_asset_uri` correctly passes `data:` URIs through, so `upload_texture` receives them legitimately — and the revoke call now fires on every `data:` image URI.

**Evidence:**
```
// gltf.rs — upload_texture onload closure
let src = img.src();
// ... texture upload ...
Url::revoke_object_url(&src).unwrap();  // BUG: called on any URI, not just blob:
```

**Action:**
- [ ] Guard the revoke call so it only fires for blob: URLs:
```rust
if src.starts_with("blob:") {
    Url::revoke_object_url(&src).unwrap();
}
```


---

---

### Response to Your Comment #1 (EXPLANATION)

> 
@wanguardd 

# AI Review Processing — gltf/file `data:` URL loader (branch: simple-filter)

Review targets commit `a88607d5` ("fix data: URL handling and error propagation in
file loader") and the surrounding branch diff vs `master`. Each comment treated as a
hypothesis and verified against the real code before acting. One commit per item.

Affected files:

- `module/helper/renderer/src/webgl/loaders/gltf.rs`
- `module/min/mingl/src/web/file.rs`

| # | Item (short) | Status | Commit | Notes |
|---|--------------|--------|--------|-------|
| 1 | Image URI loop not updated alongside buffer URI loop | **FIXED** | `fb91087a` | Real bug. Extracted `resolve_asset_uri`, used in both loops. |
| 2 | `map_err(\|_\| Other(...))` discards original error context | **FIXED (best fix, not suggested)** | `bc4292e2` | Valid. Log original error; rejected the `Other(String)` API change. |
| 3 | URL scheme detection duplicated between file.rs and gltf.rs | **FIXED (partial)** | `3c82cf61` | Scheme-list dup is real; extracted shared predicate. The `/` "divergence" is by design, not a bug. |
| 4 | No test coverage for the `data:` decode path in `load()` | **FIXED (scoped)** | `91d07993` | Valid. Extracted pure parser + 4 unit tests; rejected standing up a wasm-bindgen-test browser runner. |

## Verification

- `cargo test -p mingl --features "web,web_future,web_file" --lib` → **13 passed** (was 9; +4 new).
- `cargo check -p renderer --target wasm32-unknown-unknown` → clean after every item.

## Per-item detail

### 1 — Image URI loop not updated (FIXED, `fb91087a`)

**Real bug.** The buffer loop got the scheme-detection guard but the image loop kept
`upload_texture( format!( "{}/{}", folder_path, uri ).into() )`. A `data:`/absolute/
origin-absolute image URI was mangled (e.g. `"models/data:image/png;..."`) and silently
fell back to the 1×1 white placeholder.

**Fix (better than suggested):** extracted `resolve_asset_uri(folder_path, uri)` and call
it from **both** the buffer and image loops, so they cannot diverge again. This is the
shared-helper the reviewer suggested, scoped to the renderer crate.

### 2 — `map_err(|_| Other(...))` discards context (FIXED, `bc4292e2`)

**Valid concern.** Three sites (`load`, `from_slice`/parse, buffer `load`) discard the
underlying `JsValue` / `gltf::Error`; `Other(&'static str)` cannot carry dynamic context.

**Rejected the primary suggestion** (`Other(String)`): the variant is used at ~21 call
sites across published `min` crates, and **every** variant in `WebglError` intentionally
uses `&'static str` (house style). Widening it is a breaking API change disproportionate
to this PR.

**Applied the alternative:** log the original error (with the path) via
`gl::browser::error!` before mapping, at all three sites. Note: `gl::error!` does **not**
exist — the `error` macro name collides with the `error` *module* (the `error` layer),
which shadows the glob-imported macro; the house-style path is `gl::browser::error!`
(already used in `pmrem.rs`/`shadow.rs`). Typed error preserved for callers; full
diagnostic now reaches the console.

### 3 — Scheme detection duplicated between file.rs and gltf.rs (FIXED partial, `3c82cf61`)

**Partially valid.** The 5-scheme list (`http://`, `https://`, `//`, `blob:`, `data:`)
was genuinely duplicated between `resolve_url` (mingl) and `resolve_asset_uri` (gltf) and
could drift. Extracted a single `is_self_contained_url` predicate in `mingl::web::file`,
exposed via `mod_interface`, and called from both (`gl::file::is_self_contained_url`).

**Rejected the framing that the `/` case is a drift bug.** The leading-`/` handling
differs *by design*: `resolve_url` joins origin-absolute paths to the window origin,
whereas the gltf helper passes them through (so `file::load` resolves them later).
Folding `/` into the shared predicate would have been wrong — it stays local to each
function. Did not collapse `resolve_url` and `resolve_asset_uri` into one function: they
have distinct join rules (origin vs folder).

### 4 — No coverage for the `data:` decode path (FIXED scoped, `91d07993`)

**Valid.** The comma-detection and base64-check edge cases were exercised only in
production.

**Rejected the suggested wasm-bindgen-test + CI browser runner:** the repo has **no**
`wasm-bindgen-test` dependency or headless-browser runner anywhere in `module/min`, and
one cannot be validated in this environment. Adding unverified harness + CI config is out
of scope and risky.

**Applied the testable refactor (matches the existing `resolve_url` pattern):** extracted
the pure validation/splitting into `data_url_base64_payload(url) -> Result<&str,&'static
str>` and added 4 native `#[test]`s — valid payload, empty payload, missing comma,
non-base64. `load()` now calls the helper; only the `window.atob` byte decode (genuinely
browser-only) remains uncovered, which is stated honestly rather than faked.

## Similar-occurrence sweep

- `module/helper/renderer/src/webgl/animation/loaders/gltf.rs`: receives already-decoded
  `buffers: &[Vec<u8>]` and builds only animation-channel keys — **no** asset-URI
  concatenation, no analogous bug.
- All other `file::load` callers (`hdr_texture.rs`, examples) pass caller-supplied paths
  and now inherit the centralized scheme detection in `resolve_url` automatically.




Acknowledged in review


---

PR #198, opened 2026-06-15
NEED REVIEW : premultiplied-alpha + VertexCorners orient_to_grid / corner_source / offset / flat-tint

FairlyBet:
## Summary

Two independent improvements to `tilemap_renderer` and `tilemap_scene`, delivered in five
commits. All existing tests pass; five new tests cover the new VertexCorners paths.

---

### Problem

The WebGL adapter's `apply_blend` always used `SRC_ALPHA` as the colour source factor.
A texture exported with premultiplied alpha (RGB already scaled by alpha) drawn under
`SRC_ALPHA` is scaled by alpha a second time, producing darkened, "double-dimmed"
antialiased edges.

### What changed

**`tilemap_renderer`**

- `ImageAsset` gains `premultiplied: bool` (defaults to `false`).
- `GpuTexture` records the flag at upload time.
- `apply_blend(gl, blend, premultiplied)` gains a third parameter. When `true` the colour
  source factor becomes `ONE` (premultiplied "over"); when `false` it stays `SRC_ALPHA`.
  Affects `Normal`, `Add`, and the `Overlay`→Normal fallback.
- All three WebGL draw paths (single sprite, single mesh, batch) look up the texture's
  flag and pass it to `apply_blend`.
- SVG and terminal adapters ignore the field — no GPU blend equation involved.

**`tilemap_scene`**

- `ImageResource` (the serde-loaded RON struct) gains the same `premultiplied: bool`
  field with `#[serde(default)]` — existing RON specs need no changes.
- `compile_assets` threads the flag through to `tilemap_renderer::ImageAsset`.

### Upgrade path

No action needed for straight-alpha textures (`premultiplied: false` is the default).
Premultiplied atlas exports can opt in per-asset:

```ron
( id: "ui_atlas", path: "ui.png", premultiplied: true, ... )
```

---

## 2 — VertexCorners improvements (`tilemap_scene`)

Four orthogonal additions to `SpriteSource::VertexCorners`, all opt-in with conservative
defaults so existing specs compile unchanged.

### 2a — `orient_to_grid` — pre-baked discrete orientation

**Problem.** The legacy path bakes three 120°-rotations keyed to the canonical sort
(which slot the unique id landed in). The regular hex dual grid has **six** 60°-orientations,
so the legacy index can be off by 60° for four of them. Correcting at runtime with
`transform.rotation` would require the backend to rotate sprites around an arbitrary
pivot, adding resampling artefacts at non-right-angle angles.

**What changed.** New `orient_to_grid: bool` field (default `false`). When `true`:

- A new `dual_orientation_index` function computes the discrete orientation directly from
  the triangle's pixel geometry, aligning the *distinguishing* corner (the lone present
  corner for a corner-tile; the lone absent corner for an edge-tile) to its baked
  reference axis, then rounding to the nearest 60° step.
- `{rot}` ranges over `0..6` for corner/edge tiles and `0..2` for fully-symmetric (solid)
  tiles (only ▲/▽ parity differs for solid interiors).
- `transform.rotation` stays `0` — no runtime sprite rotation, no resampling.
- Uses `self_id` (the value in the object's all-equal `(X,X,X)` pattern, if any) to
  classify corners as "present" vs "absent". This fixes cross-region boundary orientation:
  when two different objects share a triangle edge, the old canonical-order test could
  misread the present/absent assignment and point the shape at the wrong corner.

### 2b — `corner_source` — per-layer corner channel

**Problem.** Previously every VertexCorners layer on a cell read the *terrain id* (the
highest-priority object). This made it impossible to have two independent dual grids
(e.g. base terrain + per-player region overlays) share the same scene: both would resolve
the same corner id and one would produce wrong patterns.

**What changed.** New `corner_source: Option<String>` field (default `None`):

- `None` — the cell's terrain id (historical behaviour).
- `Some("layer_name")` — the id of the first object on the cell whose `global_layer`
  matches. Each VertexCorners layer can therefore key off a distinct draw layer, keeping
  the corner lookups isolated.

New helper `tile_corner_id` in `compile/neighbors.rs`; `resolve_corners` gains a `source`
parameter threaded from each layer's sprite source.

### 2c — `offset` — world-pixel sprite shift

New `offset: Option<(f32, f32)>` field (default `None`). When set, the static `(dx, dy)`
shift is added to every emitted sprite's position. Intended for drawing the same dual grid
shifted — e.g. a grey-tinted copy nudged downward for a 2.5D wall/drop-shadow — without
disturbing corner resolution or the orient-to-grid frame pick, which always use the
unshifted triangle centre.

### 2d — `TintBehaviour::Flat` per VertexCorners layer

**Problem.** `compile_vertex_pass` hardcoded `ctx.global_tint` for every emitted sprite,
ignoring `LayerBehaviour.tint`. Per-player region overlays need independent colouring.

**What changed.** When `layer.behaviour.tint` is `Flat(TintRef)`, the named tint is
resolved via the new `resolve_tint_ref` helper (extracted from `resolve_scene_global_tint`)
and multiplied component-wise into the global tint before emitting each sprite.
`None` and `Masked` fall through to the global tint as before.

---

## Tests added

| Test | What it pins |
|---|---|
| `vertex_corners_orient_to_grid_single_hex_six_orientations` | A lone hex in void emits all 6 distinct `dual_corner_0..5` frames; `transform.rotation == 0` on each |
| `vertex_corners_orient_to_grid_up_down_distinct` | A solid-interior flower emits both `dual_full_0` (▲) and `dual_full_1` (▽) — the parity the legacy path collapsed |
| `vertex_corners_layer_flat_tint_colours_sprites` | A `Flat("red")` tint on a VertexCorners layer produces `tint ≈ [1,0,0,1]` on all emitted sprites |
| `vertex_corners_layer_no_tint_is_identity` | Default (no-tint) path still emits `[1,1,1,1]` — regression guard |
| `vertex_corners_corner_source_isolates_channels` | A cell carrying both terrain and region objects emits tiles from both assets without channel collision |

---

## Clippy

Five pedantic warnings in `tilemap_scene` cleared:

| Lint | Location | Fix |
|---|---|---|
| `single_match_else` | `frame.rs` `dual_orientation_index` | `match self_id` → `if let Some(sid)` |
| `missing_panics_doc` ×2 | `ids.rs` `alloc_image` / `alloc_sprite` | Added `# Panics` sections |
| `doc_markdown` ×2 | `load.rs`, `validate.rs` | Wrapped `RenderSpec` in backticks |

---

## Spec / roadmap

- `tilemap_scene/spec.md §4.1`: `premultiplied` added to the sampler-params block.
- `tilemap_scene/spec.md §5.6`: `orient_to_grid`, `corner_source`, `offset` documented
  with full semantics and `{rot}` range implications.
- `tilemap_scene/spec.md §11`: corner resolution rule updated to mention `corner_source`.
- `tilemap_scene/roadmap.md`: `TintBehaviour::Flat` for VertexCorners marked shipped
  (partial — `Masked` + TeamColor resolution still open). VertexCorners feature bullet
  expanded.
- `tilemap_renderer/roadmap.md`: premultiplied-alpha added to the WebGL completed list.

---

## Checklist

- [x] All pre-existing tests pass (`tilemap_renderer`, `tilemap_scene`)
- [x] 5 new tests added and passing
- [x] `cargo clippy -p tilemap_scene` — 0 warnings
- [x] Spec and roadmap updated
- [x] Backwards-compatible: new fields all have `#[serde(default)]` / `Default`


wanguardd (changes requested):
## Review Summary

| Attribute | Value |
|-----------|-------|
| Round | #1 |
| Pipeline | d284f4e |
| Rulebook | v3.19 |
| Timestamp | 2026-06-17T08:32:37Z |
| Order |  |

---

System of Review completed analysis of 'NEED REVIEW : premultiplied-alpha + VertexCorners orient_to_grid / corner_source / offset / flat-tint'. Found **14 findings** across 4 files.

## 📋 Review Summary

| 🐛 CORRECTNESS | 3 | 0 | 3 |
| 📋 REQUIREMENTS | 1 | 2 | 3 |
| 📐 DESIGN | 0 | 1 | 1 |
| 🔍 QUALITY | 0 | 1 | 1 |
| ✅ TESTING | 0 | 4 | 4 |
| 📖 DOCUMENTATION | 0 | 2 | 2 |
| **Total** | **4** | **10** | **14** |

---

## 🐛 CORRECTNESS

🔴 INTRODUCED: Wildcard pattern ("*","*","*") with orient_to_grid: true causes CompileError::UnresolvedRef at runtime. The frame pre-allocation in assets.rs evaluates `a == b && b == c` to pick 2 or 6 frames; since "*"=="*" is true, a wildcard pattern gets only 2 frames. At runtime dual_orientation_index() with self_id=None (wildcards are excluded from self_id by the `p.corners.0 != "*"` guard) uses the legacy canonical-sort path. For triangles where the three corner ids differ, this path sets period=6 and returns rot_index 2–5 — referencing frames that were never allocated, causing CompileError::UnresolvedRef for any scene using ("*","*","*") with orient_to_grid: true.

**Evidence:**
```json
assets.rs:207: `if a == b && b == c { 2_u32 } else { 6 }` — no wildcard guard; "*"=="*" triggers 2-frame path. frame.rs:329: self_id guard `p.corners.0 != "*"` excludes wildcards from self_id → self_id=None. frame.rs:232-239: legacy path with period=6 returns steps.rem_euclid(6) in 0–5. frame.rs:338-345: CompileError::UnresolvedRef fires for any rot_index ≥ 2 on 2-frame pattern.
```

**Action:**
- [ ] In assets.rs:207, change the condition to `a == b && b == c && a.as_str() != "*"` so wildcard patterns always allocate 6 frames in orient_to_grid mode. (Self-id detection already correctly excludes wildcards, so frame.rs needs no change.)


---

🔴 INTRODUCED: TintBehaviour::Flat is silently ignored on non-VertexCorners layers. compile_vertex_pass() correctly handles Flat (frame.rs:360-370), but compile_instance_layer() — used for hex, edge, free, and viewport layers — calls final_tint(ctx.global_tint, layer.behaviour.alpha, inst.tint) without reading layer.behaviour.tint. final_tint() has no parameter for a named tint override. A layer with tint: Flat("#ff0000") on a regular terrain layer emits sprites tinted only by the global tint — the red tint is silently discarded.

**Evidence:**
```rust
frame.rs:360-370: compile_vertex_pass Flat arm applies global_tint * c. frame.rs:1003,1038,1155,1267,1354,1387,1482,1503: compile_instance_layer calls final_tint() — no layer.behaviour.tint access. frame.rs:1167: fn final_tint(global, layer_alpha, inst) — no tint field parameter. Spec §6.1 implies Flat applies to any layer type.
```

**Action:**
- [ ] Either (a) implement Flat tint in compile_instance_layer — read layer.behaviour.tint before calling final_tint and multiply the result when Flat — or (b) restrict Flat to VertexCorners by emitting a new CompileError::UnsupportedBehaviour for Flat on other layer types, and update spec §6.1 to document the restriction.


---

🔴 INTRODUCED: TintBehaviour::Masked silently falls through to global_tint on VertexCorners layers. The match arm `_ => ctx.global_tint` at frame.rs:370 catches Masked, discarding the mask sprite reference and tint field with no error or warning. Spec §6.1 specifies Masked as sampling a second sprite for per-pixel tinting.

**Evidence:**
```json
frame.rs:358-370: match &layer.behaviour.tint { TintBehaviour::Flat(tref) => { ... }, _ => ctx.global_tint } — Masked hits the fallthrough arm. No CompileError variant for unsupported tint behaviour. Spec §6.1: "Masked { mask, tint } — multiply the sprite by a mask texture sampled per-pixel."
```

**Action:**
- [ ] Add a CompileError::UnsupportedBehaviour variant (or reuse an existing error type) and emit it from compile_vertex_pass when layer.behaviour.tint is Masked, with message: "Masked tint not yet implemented for VertexCorners — use Flat or remove the tint behaviour." Document in spec §6.1 and roadmap that Masked is unimplemented for VertexCorners.


---

## 📋 REQUIREMENTS

🔴 INTRODUCED: TriBlendPattern.sprite_pattern doc comment says legacy {rot} ranges over "0..2" (2 values) but implementation allocates 3 frames and canonicalize() returns rotation in {0,1,2}. An atlas author following the struct doc would create 2 legacy frames and hit a missing-sprite error when rotation==2.

**Evidence:**
```json
source.rs:252: "In legacy mode the index ranges over `0..2` (canonical-sort rotation)" — "0..2" in Rust notation means values {0,1} only. source.rs:253: second occurrence of "0..2" for the symmetric case. spec.md:487: "When false (legacy default), `{rot}` ranges over `0..3`" — spec is correct. assets.rs:210: `for rot in 0_u32..count` where count=3 for legacy mode — allocates {0,1,2}.
```

**Action:**
- [ ] Fix source.rs lines 252 and 253: change "0..2 (canonical-sort rotation)" to "0..3 (canonical-sort rotations 0, 1, 2)" to match the spec and implementation.


---

🟠 AFFECTED: Spec §11 uses ambiguous "lowest-indexed object id" for corner resolution but the implementation iterates tile.objects in declaration order and returns the first with any priority — not the lowest priority value, not the lexicographically smallest id string.

**Evidence:**
```json
spec.md:917: "lowest-indexed object id on the cell that has a priority field set" — "lowest-indexed" is ambiguous. neighbors.rs:218-237: tile_terrain_id() uses `for object_id in &tile.objects { ... return Some(object_id) }` — declaration-order first. neighbors.rs:241-244: tile_corner_id() doc: "first object with a priority" — the clearer language used in the new function.
```

**Action:**
- [ ] Update spec.md §11 to: "the id of the first object in the tile's object list (in declaration order) that has a priority field set" — matching the implementation semantics and the tile_corner_id() doc.


---

📋 MINOR: Four #[allow] attributes in PR-modified functions lack required explanatory comments per project rulebook.md line 98: "Each allow attribute should have a short comment explaining why it is needed."

**Evidence:**
```
rulebook.md:98: "Each allow attribute should have a short comment explaining why it is needed." vertex.rs:87: `#[ allow( clippy::implicit_hasher ) ]` on resolve_corners (signature modified by this PR). vertex.rs:114: `#[ allow( clippy::needless_pass_by_value ) ]` on canonicalize (doc modified by this PR). neighbors.rs:135,185: `#[ allow( clippy::implicit_hasher ) ]` on compute_neighbor_bitmask and neighbor_state_at — no explanatory comments on any of these four.
```

**Action:**
- [ ] Add a short inline comment after each #[allow]: e.g., for implicit_hasher: "// caller's HashMap type is sufficient — generic hasher not needed" and for needless_pass_by_value: "// takes [String;3] by value — sort-in-place semantics are intentional."


---

## 📐 DESIGN

📐 MINOR: A misconfigured corner_source layer name silently degrades all corners to VOID_ID with no warning or error. tile_corner_id() returns None when no object on the cell matches the named global_layer, and resolve_corners() translates None to VOID_ID. A typo like corner_source: Some("Terrain") vs "terrain" produces all-void rendering with no diagnostic.

**Evidence:**
```
neighbors.rs:248: tile_corner_id() returns None on no match. compile/vertex.rs: resolve_corners() maps None → VOID_ID. validate.rs: no validation for corner_source layer name against scene objects. spec.md §5.6: no mention of this silent fallback.
```

**Action:**
- [ ] Document the fallback in spec.md §5.6 after the corner_source description: "If no object on the cell matches the named layer, the corner resolves to VOID_ID — verify layer names carefully." Optionally, add a validation warning when a VertexCorners layer's corner_source matches no object global_layer in the entire scene.


---

## 🔍 QUALITY

🔍 MINOR: The depth-sort key for VertexCorners sprites uses the offset-adjusted world position (wx+ox, wy+oy) rather than the true geometric centroid. Sprites using offset: Some((dx,dy)) will shift their z-order relative to non-offset sprites, potentially causing unexpected occlusion for 2.5D shadow use-cases. This behavior is undocumented.

**Evidence:**
```json
frame.rs:349-350: `let (ox,oy) = offset.unwrap_or((0.0,0.0)); let (wx,wy) = (wx+ox, wy+oy)` — offset applied before (wx,wy) is used as position/sort key. spec.md §5.6 offset description: "corner resolution and orient-to-grid frame selection always use the true (un-shifted) triangle geometry" — does not mention sort key behavior.
```

**Action:**
- [ ] Document in the offset field doc comment (resource.rs or source.rs) that offset affects the depth-sort key, so that atlas authors using offset for 2.5D shadows understand the z-ordering consequence. If the intent is that offset should NOT affect sort order, apply offset only after the sort key is computed.


---

## ✅ TESTING

✅ MAJOR: The offset field (Option<(f32,f32)>) has zero test coverage. No test in the PR sets offset: Some((dx,dy)). Neither the position-shift behavior nor the claimed invariant that "corner resolution and orient-to-grid frame selection always use the true (un-shifted) triangle geometry" are tested.

**Evidence:**
```
grep -rn 'offset.*Some\|Some.*offset' in tilemap_scene/tests/ — no matches. All 5 new tests in scene_model_compile_test.rs use offset: None or omit the field. frame.rs:349-350: straightforward implementation but completely untested.
```

**Action:**
- [ ] Add a test that creates a VertexCorners layer with offset: Some((32.0, -16.0)), compiles the scene, and asserts the resulting sprite position differs from the no-offset baseline by exactly (32.0, -16.0).


---

✅ MAJOR: The premultiplied: true path has zero test coverage. All fixture uses across 6 test files use premultiplied: false (the serde default). No test verifies that premultiplied: true propagates through compile_assets into the ImageAsset.premultiplied field.

**Evidence:**
```json
grep -rn 'premultiplied.*true' in tilemap_scene/tests/ — no matches. 15+ fixture sites across scene_model_compile_test.rs, scene_model_test.rs, and other test files all use premultiplied: false or rely on #[serde(default)].
```

**Action:**
- [ ] Add a test: create ImageResource with premultiplied: true, call compile_assets, assert compiled.assets.images[0].premultiplied == true.


---

✅ MAJOR: The cross-region boundary scenario that motivated the self_id fix has no test. frame.rs:182-190 doc comment explicitly describes: "for region_1's edge tile the corners are (region_1, region_1, region_0) and the old canonical-order test misread the edge as a corner." None of the 5 new tests create two adjacent hexes with different region objects and orient_to_grid: true.

**Evidence:**
```
grep -rn 'adjacent\|region_0.*region_1\|cross.region' in tilemap_scene/tests/scene_model_compile_test.rs — no matches. All new orient_to_grid tests use single-object flower scenes.
```

**Action:**
- [ ] Add a two-hex scene test where each hex has a different region object, both with orient_to_grid: true. Assert that the shared-edge triangle emits the geometrically correct orientation frame for each region object (verifying the self_id fix prevents canonical-sort misclassification).


---

✅ MINOR: vertex_corners_corner_source_isolates_channels proves presence of both asset families but not isolation between channels. A bug where corner_source is silently ignored would also pass — both asset families happen to match from the same default terrain channel.

**Evidence:**
```
scene_model_compile_test.rs:1758-1763: asserts `any_dual && any_region` — only presence. No assertion that dual-layer sprites contain no region_* frames, or that region-layer sprites contain no dual_corner_* frames.
```

**Action:**
- [ ] Add exclusion assertions: for the dual layer, assert that no sprite frame name contains "region_"; for the region layer, assert that no sprite frame name contains "dual_corner_".


---

## 📖 DOCUMENTATION

📖 MAJOR: Spec §5.6 documents the orient_to_grid frame ranges (0..2 / 0..6) but omits the baker contract needed to produce a compatible atlas. The frame.rs dual_orientation_index() doc comment (lines 171-175) contains the full contract — baker CCW rotation appears as CW in world due to PNG Y-flip, reference corner points at base − 60°·o — but this information is absent from the spec. An atlas artist cannot correctly order frames from the spec alone.

**Evidence:**
```
grep -n 'baker\|CCW\|clockwise\|orientation 0' module/helper/tilemap_scene/spec.md — no matches. frame.rs:171-175: complete baker contract in implementation doc only.
```

**Action:**
- [ ] Add a "Baker contract" subsection to spec.md §5.6 under the orient_to_grid bullet: "(a) orientation 0 is defined by the base angle — 300° for edge tiles, 60° for corner tiles; (b) {rot} increments clockwise in world space (the atlas baker's CCW rotation appears CW after PNG vertical flip); (c) the reference corner for each frame points at base − 60°·rot in world space."


---

📖 MINOR: Spec §5.6 and the VertexCorners orient_to_grid field doc both state the 2-frame symmetric case triggers when "all three corners equal", but the code uses present-corner counting against self_id. At cross-region boundaries the two criteria diverge: a triangle with (region_1,region_1,region_0) has all-different canonical corners (not "all equal") but has 2 of 3 corners present for region_1 → period=6 in the implementation.

**Evidence:**
```
spec.md:487: "all three corners equal — only ▲/▽ parity distinguishes them." source.rs:89: same "all three corners equal" language. frame.rs:206-226: actual logic uses present-corner counting against self_id — 3 or 0 present → period=2; 2 present → period=6; 1 present → period=6.
```

**Action:**
- [ ] Update spec.md §5.6 and source.rs orient_to_grid field doc to describe the actual mechanism: "a triangle is fully-symmetric (2-frame) when all 3 corners belong to this object (solid interior) or none do (degenerate void) — otherwise it is asymmetric (6-frame)."


---

---

>
>

FairlyBet:
@wanguardd 

# AI Review Processing — tilemap_scene (branch: fix-warnings)

Each item treated as a hypothesis. Every fix verified by test or reasoning; full
crate suite green (181 tests across all binaries). One commit per item.

| # | Item (short) | Status | Commit | Notes |
|---|--------------|--------|--------|-------|
| 1 | Wildcard `("*","*","*")` + orient_to_grid → 2 frames, runtime needs 6 | **FIXED** | a53123aa | Real bug. Added `!= "*"` guard. Regression test verified failing pre-fix. |
| 2 | `Flat` tint ignored on non-VertexCorners layers | **FIXED** | 18afb385 | Real (spec §6.1 is generic). Added `layer_base_tint` helper, applied across all emit paths. |
| 3 | `Masked` tint silently → global in vertex pass | **FIXED** | d3264097 | Real. Added `CompileError::UnsupportedBehaviour`; exhaustive match. |
| 4 | source.rs doc `{rot}` legacy "0..2" | **FIXED (partial)** | d80683ed | Legacy "0..2"→"0..3" fixed. **Rejected** the suggested edit to the symmetric "0..2" (that one is correct: period 2 = {0,1}). |
| 5 | spec §11 "lowest-indexed object id" ambiguous | **FIXED** | d4dbc272 | Reworded to "first object in declaration order with a priority". |
| 6 | Four `#[allow]` lack comments | **FIXED** | e118ed83 | Fixed the 4 flagged + 3 other uncommented item-level allows in the crate (same rule). |
| 7 | corner_source typo silently → VOID_ID | **FIXED (doc only)** | f42bc919 | Documented fallback. **Rejected** the optional validation (a no-match channel is legitimate; would reject valid scenes, and no warning channel exists). |
| 8 | offset affects depth-sort key (undocumented) | **FIXED (doc)** | 0e738107 | Documented current/intended behaviour in the offset field doc. |
| 9 | offset field zero test coverage | **FIXED** | 8b7fa199 | Added test: position delta == offset, same frames (un-shifted geometry invariant). |
| 10 | premultiplied:true zero coverage | **FIXED** | 7fd9ba2c | Added propagation test (true propagates, default false). |
| 11 | cross-region boundary self_id fix untested | **FIXED** | 13b4c117 | Added invariant test (foreign region orients identically to void). Verified failing when self_id forced to None. |
| 12 | channel-isolation test only proves presence | **FIXED** | 7a01e4f4 | Added cross-family exclusion assertions (dual→only dual_corner_*, region→only r_corner_*). |
| 13 | spec §5.6 missing baker contract | **FIXED** | d50b873c | Added "Baker contract" subsection mirroring the impl doc. |
| 14 | spec §5.6 / source.rs "all corners equal" wording | **FIXED** | 2bd76a6e | Reworded to self-id counting mechanism in both spec and field doc. |

## Verdicts on disputed/partial items

- **#4 — partially rejected.** The reviewer asked to change *both* `0..2` strings to
  `0..3`. Only the **legacy** one is wrong (canonicalize returns {0,1,2}, 3 frames).
  The second `0..2` describes the orient-mode fully-symmetric case, which really is
  2 frames (period 2 = values {0,1}) and matches spec §5.6 — changing it would have
  made the doc wrong. Fixed legacy only.

- **#7 — validation rejected, doc done.** Documenting the silent fallback is correct.
  The *optional* validation warning was not added: a `corner_source` that matches no
  object on a given cell is a legitimate outcome (that corner is simply void), so a
  hard error would reject valid scenes, and `validate.rs` has only an error channel,
  no warning channel.

- **#2 — chose implement over restrict.** Reviewer offered (a) implement Flat
  everywhere or (b) restrict Flat to VertexCorners + error. Chose (a): spec §6.1
  defines Flat/Masked generically and the spec's own example (§ knight, line 808)
  applies Masked to an instance layer, so tint behaviours are clearly meant for all
  layers. Restricting would contradict the spec.

All other items were valid and fixed as the best available fix (some refactored for
shared code rather than copy-paste, e.g. the `layer_base_tint` helper unifying #2/#3).

wanguardd (changes requested):
## 🐛 CORRECTNESS

🔴 INTRODUCED — webgl_helpers.rs:698 BlendMode::Multiply ignores premultiplied flag; src_a computed but not used, causing double-alpha darkening for premultiplied textures

**Evidence:**
```json
webgl_helpers.rs:683: src_a = if premultiplied { gl::ONE } else { gl::SRC_ALPHA }. Line 698 Multiply arm: blend_func_separate(gl::DST_COLOR, gl::ONE_MINUS_SRC_ALPHA, ...) — uses hard-coded gl::DST_COLOR, not src_a. Normal/Add arms at lines 691/725 correctly pass src_a as the first argument. For a premultiplied texture, src.rgb already contains original_rgb*src.a; using DST_COLOR (not src_a) multiplies alpha a second time on semi-transparent edges.
```

**Action:**
- [ ] Replace gl::DST_COLOR with src_a in the BlendMode::Multiply blend_func_separate call at webgl_helpers.rs:698, matching the fix already applied to Normal and Add. If Multiply is intentionally restricted to straight-alpha, add a compile-time reject for premultiplied=true textures with BlendMode::Multiply and document the restriction.


---

🔴 INTRODUCED — frame.rs:327-329 self_id extracted from first all-equal pattern; second-terrain triangles silently receive orientation index 0

**Evidence:**
```
frame.rs:327-329: self_id = patterns.iter().find_map(|p| (p.corners.0==p.corners.1 && p.corners.1==p.corners.2 && p.corners.0!="*").then_some(p.corners.0.as_str())). When patterns list contains ("grass","grass","grass") first and ("water","water","water") second, self_id is always "grass". Water-only triangles (all corners=="water") have count=0 relative to self_id="grass", fall to the _ arm with period=2, and always emit orientation 0. Every water interior triangle renders the same frame with no error.
```

**Action:**
- [ ] Compute self_id per-triangle from the canonical corners directly: if all three canonical corners are equal and non-wildcard, use that value as self_id for this triangle rather than scanning the full pattern list. This matches the intended semantics of "the triangle's own terrain id" and eliminates the first-match bias.


---

🟡 INTRODUCED — webgl.rs:750 mesh batch draw hardcodes premultiplied=false while webgl.rs:373 single-command DrawMesh reads premultiplied from the texture, producing inconsistent blend for textured mesh batches

**Evidence:**
```rust
webgl.rs:750: GpuBatch::Mesh { params, .. } => (&params.blend, false) — hardcodes false. webgl.rs:373: premultiplied = m.texture.and_then(|id| res.texture(id)).map_or(false, |t| t.premultiplied) — reads from texture. commands.rs MeshBatchParams has pub texture: Option<ResourceId<asset::Image>>, confirming a mesh batch can carry a premultiplied texture. Comment at webgl.rs:744-745 asserts "mesh batches are straight-alpha" but this is undocumented in the MeshBatchParams API and contradicted by the single-command path's behavior.
```

**Action:**
- [ ] Either read premultiplied from the batch texture for consistency with the single-command path, or document the MeshBatchParams restriction explicitly and add a validation guard at CreateMeshBatch time that rejects premultiplied=true textures assigned to mesh batches.


---

## 🔍 QUALITY

🟡 INTRODUCED — assets.rs:208-209 comment misattributes the != "*" wildcard guard to dual_orientation_index; the guard lives in compile_vertex_pass

**Evidence:**
```
assets.rs:208-209: 'A ("*","*","*") pattern is excluded from self_id detection by the != "*" guard in dual_orientation_index'. grep frame.rs:327-329 shows the guard p.corners.0 != "*" is inside compile_vertex_pass's find_map closure, not inside dual_orientation_index (frame.rs:166-247). dual_orientation_index receives self_id as a parameter and contains no guard of its own.
```

**Action:**
- [ ] Correct assets.rs:208-209 to: "...excluded from self_id detection by the != '*' guard in compile_vertex_pass (see frame.rs:328)." This prevents future maintainers from searching in the wrong function when tracing the wildcard-exclusion logic.


---

## ✅ TESTING

🔴 INTRODUCED — scene_model_compile_test.rs has zero orient_to_grid tests for HexPointyTop tiling; the pointy-top coordinate dispatch branch is entirely uncovered

**Evidence:**
```
grep -n 'HexPointyTop' scene_model_compile_test.rs returns exit 1 (zero matches). All three orient_to_grid tests (vertex_corners_orient_to_grid_single_hex_six_orientations, _up_down_distinct, _triple_wildcard_allocates_six) call dual_orient_spec() which uses TilingStrategy::HexFlatTop. The enumerate_triangles function dispatches to different coordinate types for HexPointyTop vs HexFlatTop — a bug in the pointy-top path would not be caught.
```

**Action:**
- [ ] Add vertex_corners_orient_to_grid_pointy_top_six_orientations test in scene_model_compile_test.rs: construct a single-hex spec with TilingStrategy::HexPointyTop and orient_to_grid:true, assert that exactly 6 distinct {rot} frame indices appear in the emitted sprites for a hex surrounded by void.


---

🔴 INTRODUCED — scene_model_compile_test.rs has no test for corner_source pointing to a nonexistent layer; the documented silent VOID_ID fallback is untested

**Evidence:**
```
grep -n 'corner_source' scene_model_compile_test.rs shows only corner_source:None (lines 1332,1435,1536,1680) and corner_source:Some("region") — a valid layer name (lines 1958,2083). spec.md §5.6 explicitly documents the silent fallback: "a misspelled corner_source names no object's global_layer, so every corner resolves to void with no error." No test covers this path; a regression converting it to a CompileError or panic would go undetected.
```

**Action:**
- [ ] Add vertex_corners_corner_source_invalid_layer_falls_back_to_void: set corner_source:Some("no_such_layer") on a hex scene, assert that the emitted frame names match what corner_source:None produces for the same geometry (all void-edge frames), confirming the silent fallback is preserved.


---

## 📖 DOCUMENTATION

🟡 INTRODUCED — roadmap.md:141-145 misrepresents Flat tint scope as compile_vertex_pass-only; implementation is universal via the shared layer_base_tint helper

**Evidence:**
```
roadmap.md:142: "compile_vertex_pass now respects LayerBehaviour.tint = Flat(TintRef) per object layer". grep frame.rs for layer_base_tint shows call sites at lines 358, 990, 1025, 1142, 1292, 1379, 1412, 1507 — spanning compile_vertex_pass, compile_instance_layer, edge pass, viewport pass, and free pass. Flat tint applies universally to all layer types, not only to VertexCorners.
```

**Action:**
- [ ] Update roadmap.md item 1: replace "compile_vertex_pass now respects..." with "TintBehaviour::Flat is now implemented for all layer types via the shared layer_base_tint helper. TintBehaviour::Masked is explicitly rejected with CompileError::UnsupportedBehaviour."


---

🟡 INTRODUCED — spec.md §5.6 is silent on the three-id chiral junction limitation; atlas authors cannot know that all-distinct-corner triangles fall back to a non-geometric orientation

**Evidence:**
```
frame.rs:192-194 code comment: "still assumes at most two distinct ids per triangle drive one object shape. A genuine three-id chiral junction triangle is out of scope." grep spec.md for 'chiral', 'three-id', 'distinct.*corner', 'out of scope' returns no results in the §5.6 baker contract section. An author following only spec.md has no indication this constraint exists or what output to expect.
```

**Action:**
- [ ] Add a note to spec.md §5.6 baker contract: "Three-id chiral junctions (all three corners hold distinct non-void ids) are out of scope for orient_to_grid in v0.2.0. The emitted frame is deterministic but not geometrically meaningful. Suppress these triangles by providing a more-specific pattern."


---

🟡 INTRODUCED — spec.md:928 uses "declaration order" for terrain id resolution but this ordering is undefined for dynamically-constructed scenes where slotmap iteration order is used

**Evidence:**
```
spec.md:928: "the first object in the cell's object list (in declaration order) that has a priority field set". neighbors.rs:228-237 iterates tile.objects populated from hex_instances() which uses slotmap iteration order — deterministic but not guaranteed to match spec declaration order in dynamically built scenes. For statically parsed scenes the orders coincide; for API-constructed scenes the spec leaves the tiebreak undefined.
```

**Action:**
- [ ] Clarify spec.md §11 to distinguish static "declaration order" from dynamic "insertion order". Either specify that insertion order must match declaration order (banning post-creation reordering), or state explicitly that API-constructed scenes resolve ties by insertion order and spec authors should avoid relying on object ordering within a cell.


---

---

### Response to Your Comment #1 (EXPLANATION)

> 
@wanguardd 

# AI Review Processing — tilemap_scene (branch: fix-warnings)

Each item treated as a hypothesis. Every fix verified by test or reasoning; full
crate suite green (181 tests across all binaries). One commit per item.

| # | Item (short) | Status | Commit | Notes |
|---|--------------|--------|--------|-------|
| 1 | Wildcard `("*","*","*")` + orient_to_grid → 2 frames, runtime needs 6 | **FIXED** | a53123aa | Real bug. Added `!= "*"` guard. Regression test verified failing pre-fix. |
| 2 | `Flat` tint ignored on non-VertexCorners layers | **FIXED** | 18afb385 | Real (spec §6.1 is generic). Added `layer_base_tint` helper, applied across all emit paths. |
| 3 | `Masked` tint silently → global in vertex pass | **FIXED** | d3264097 | Real. Added `CompileError::UnsupportedBehaviour`; exhaustive match. |
| 4 | source.rs doc `{rot}` legacy "0..2" | **FIXED (partial)** | d80683ed | Legacy "0..2"→"0..3" fixed. **Rejected** the suggested edit to the symmetric "0..2" (that one is correct: period 2 = {0,1}). |
| 5 | spec §11 "lowest-indexed object id" ambiguous | **FIXED** | d4dbc272 | Reworded to "first object in declaration order with a priority". |
| 6 | Four `#[allow]` lack comments | **FIXED** | e118ed83 | Fixed the 4 flagged + 3 other uncommented item-level allows in the crate (same rule). |
| 7 | corner_source typo silently → VOID_ID | **FIXED (doc only)** | f42bc919 | Documented fallback. **Rejected** the optional validation (a no-match channel is legitimate; would reject valid scenes, and no warning channel exists). |
| 8 | offset affects depth-sort key (undocumented) | **FIXED (doc)** | 0e738107 | Documented current/intended behaviour in the offset field doc. |
| 9 | offset field zero test coverage | **FIXED** | 8b7fa199 | Added test: position delta == offset, same frames (un-shifted geometry invariant). |
| 10 | premultiplied:true zero coverage | **FIXED** | 7fd9ba2c | Added propagation test (true propagates, default false). |
| 11 | cross-region boundary self_id fix untested | **FIXED** | 13b4c117 | Added invariant test (foreign region orients identically to void). Verified failing when self_id forced to None. |
| 12 | channel-isolation test only proves presence | **FIXED** | 7a01e4f4 | Added cross-family exclusion assertions (dual→only dual_corner_*, region→only r_corner_*). |
| 13 | spec §5.6 missing baker contract | **FIXED** | d50b873c | Added "Baker contract" subsection mirroring the impl doc. |
| 14 | spec §5.6 / source.rs "all corners equal" wording | **FIXED** | 2bd76a6e | Reworded to self-id counting mechanism in both spec and field doc. |

## Verdicts on disputed/partial items

- **#4 — partially rejected.** The reviewer asked to change *both* `0..2` strings to
  `0..3`. Only the **legacy** one is wrong (canonicalize returns {0,1,2}, 3 frames).
  The second `0..2` describes the orient-mode fully-symmetric case, which really is
  2 frames (period 2 = values {0,1}) and matches spec §5.6 — changing it would have
  made the doc wrong. Fixed legacy only.

- **#7 — validation rejected, doc done.** Documenting the silent fallback is correct.
  The *optional* validation warning was not added: a `corner_source` that matches no
  object on a given cell is a legitimate outcome (that corner is simply void), so a
  hard error would reject valid scenes, and `validate.rs` has only an error channel,
  no warning channel.

- **#2 — chose implement over restrict.** Reviewer offered (a) implement Flat
  everywhere or (b) restrict Flat to VertexCorners + error. Chose (a): spec §6.1
  defines Flat/Masked generically and the spec's own example (§ knight, line 808)
  applies Masked to an instance layer, so tint behaviours are clearly meant for all
  layers. Restricting would contradict the spec.

All other items were valid and fixed as the best available fix (some refactored for
shared code rather than copy-paste, e.g. the `layer_base_tint` helper unifying #2/#3).




Acknowledged in review


---

FairlyBet:
@wanguardd 

# Code Review Processing Log

Branch: `fix-warnings`. Each item is treated as a hypothesis and verified
against the code before acting. One commit per fixed item; rejections carry no
commit but record the reasoning. No co-author trailers.

| # | Review item | Status | Commit | Explanation |
|---|-------------|--------|--------|-------------|
| 1 | webgl_helpers.rs:698 `Multiply` ignores premultiplied flag | **Rejected** | — | Suggested fix breaks Multiply; premultiplied path already correct. See below. |
| 2 | frame.rs:327-329 self_id first-match bias → orientation 0 | **Rejected** | — | Described impact false; suggested fix regresses cross-region fix. See below. |
| 3 | webgl.rs:750 mesh batch hardcodes premultiplied=false | **Fixed** | `6786867e` | Real inconsistency: mesh batch can carry a premultiplied texture. Now reads it from the batch texture, matching the single-command path. |
| 4 | assets.rs:208-209 comment misattributes `!= "*"` guard | **Fixed** | `fdd23756` | Guard lives in `compile_vertex_pass` (frame.rs), not `dual_orientation_index`. Comment repointed. |
| 5 | No HexPointyTop orient_to_grid test | **Fixed** | `76c22e63` | Added test; it surfaced a real bug — pointy-top bearings are 30° off, hitting `round()` boundaries. Fixed the discretization (30° base offset for pointy) and added the test. |
| 6 | No corner_source invalid-layer fallback test | **Fixed** | `6384d67d` | Added a test pinning the documented silent void fallback (typo'd channel → zero region frames). |
| 7 | roadmap.md:142 Flat tint scope misrepresented | **Fixed** | `88187d4e` | Flat tint is universal via `layer_base_tint`, not VertexCorners-only; noted Masked is rejected. |
| 8 | spec.md §5.6 silent on three-id chiral junction | **Fixed** | `864308dd` | Documented the out-of-scope three-distinct-id junction limitation in the baker contract. |
| 9 | spec.md:928 "declaration order" undefined for dynamic scenes | **Fixed** | `f7077bad` | Clarified the tiebreak is the cell's object-list order (= declaration order only for parsed scenes; insertion order otherwise). |

**Summary:** 7 fixed, 2 rejected. Rejections (items 1 and 2) were both
`🔴 CORRECTNESS` claims whose suggested fixes were actively harmful. Item 5
(a `🔴 TESTING` item) was the highest-value one — adding the requested coverage
exposed a genuine correctness bug in the pointy-top orientation path.

---

## Item 1 — REJECTED

**Claim:** `BlendMode::Multiply` uses hard-coded `gl::DST_COLOR` instead of
`src_a`, causing double-alpha darkening for premultiplied textures. Suggested
fix: replace `gl::DST_COLOR` with `src_a`.

**Verification:** The suggested fix is wrong and would destroy Multiply. The
source-color factor for Multiply *must* be `DST_COLOR` — that is what produces
`src·dst`. Replacing it with `src_a` (`ONE`/`SRC_ALPHA`) yields
`src·src_a + dst·(1-src_a)`, i.e. plain Normal blending, not Multiply.

For a **premultiplied** source the current code is already correct. With opaque
dst the W3C "over" composite of a Multiply blend is
`αs·Cs·Cb + (1-αs)·Cb`. Premultiplied `src.rgb = αs·Cs`, so
`src.rgb·DST_COLOR + dst·(1-αs) = αs·Cs·Cb + Cb·(1-αs)` — an exact match. The
`src_a` factor is deliberately unused here because `DST_COLOR` carries no source
alpha; premultiplication is already baked into `src.rgb`.

The only real approximation is for **straight-alpha** sources with `αs<1`, and
that is *already documented* in the comment at lines 692-696. The reviewer
inverted which path is affected. Rejected.

## Item 2 — REJECTED

**Claim:** `self_id` is taken from the first all-equal pattern, so a layer with
`(grass,grass,grass)` then `(water,water,water)` always resolves `self_id =
"grass"`; water interior triangles get count 0 and "always emit orientation 0".
Suggested fix: derive `self_id` per-triangle from the canonical corners (all
three equal → that id).

**Verification:** Two problems.

1. The described impact is false. A `(water,water,water)` triangle gives
   `count == 0`, which falls into the same `_ => (FRAC_PI_3, 2, 0)` arm as a
   full tile (`count == 3`): period **2**, parity computed from geometry — not a
   constant 0. A solid water triangle is correctly rendered as ▲/▽ parity.

2. The suggested fix regresses the cross-region orientation fix landed in
   13b4c117. For an edge triangle `(region_1, region_1, region_0)` the canonical
   corners are *not* all equal, so per-triangle derivation yields `None` and
   falls back to the legacy canonical-sort rule — exactly the bug that fix
   removed (`"region_0" < "region_1"` misreads the edge as a corner).

The genuine kernel — one layer carrying two distinct non-wildcard `(X,X,X)`
patterns — is outside the format model: spec §5.6 and the `dual_orientation_index`
doc both define the self id as singular ("this layer's own object id"). No
spec'd or tested configuration hits it. Rejected; the harmful fix is not applied.

wanguardd (changes requested):
## 🐛 CORRECTNESS

🔴 INTRODUCED — Build failure: unknown

**Evidence:**
```
Build exit code 1; see 024_build_output.log for details
```

**Action:**
- [ ] Fix all build errors before merging


---

🔴 INTRODUCED — `ImageAsset` at module/helper/tilemap_renderer/src/assets.rs:174 lacks `Default` impl; new `premultiplied: bool` field breaks any downstream code constructing it with an exhaustive struct literal

**Evidence:**
```rust
grep -n 'impl Default|non_exhaustive|derive.*Default' module/helper/tilemap_renderer/src/assets.rs returns empty; struct derives only Debug; premultiplied: bool added at line 195; lib.rs has #\![ allow( clippy::exhaustive_structs ) ] which explicitly opts into exhaustive struct semantics
```

**Action:**
- [ ] Add impl Default for ImageAsset with premultiplied: false, or add #[non_exhaustive] to the struct definition


---

🔴 RE-REQUEST — `BlendMode::Multiply` in module/helper/tilemap_renderer/src/adapters/webgl/webgl_helpers.rs:698 ignores the premultiplied flag; src_a is computed at line 687 but not used in the Multiply arm, causing double-alpha darkening for premultiplied textures

**Evidence:**
```json
webgl_helpers.rs:687: let src_a = if premultiplied { gl::ONE } else { gl::SRC_ALPHA }; webgl_helpers.rs:698: BlendMode::Multiply => gl.blend_func_separate( gl::DST_COLOR, gl::ONE_MINUS_SRC_ALPHA, gl::ONE, gl::ONE_MINUS_SRC_ALPHA ) — src_a not used; contrast with Add (line 691) and Normal (line 725) which correctly use src_a
```

**Action:**
- [ ] Replace gl::DST_COLOR with src_a in the BlendMode::Multiply arm, or document that Multiply does not support premultiplied textures


---

## 📐 DESIGN

🟡 INTRODUCED — corner_source in module/helper/tilemap_scene/src/compile/neighbors.rs:258-270 has no compile-time validation; a misspelled layer name silently returns VOID_ID for all corners with no error or warning, producing geometrically incorrect output

**Evidence:**
```
neighbors.rs:258-270 tile_corner_id performs runtime-only match (obj.global_layer == layer); grep -n 'corner_source' module/helper/tilemap_scene/src/validate.rs returns no matches — no compile-phase check; test vertex_corners_corner_source_invalid_layer_falls_back_to_void confirms silent VOID_ID fallback is the only behavior
```

**Action:**
- [ ] Add an optional compile-phase warning in compile_assets that checks corner_source values against the set of known global_layer names in the spec


---

## ✅ TESTING

🟡 INTRODUCED — No regression test for mesh-batch premultiplied fix (commit 6786867e) in module/helper/tilemap_renderer/src/adapters/webgl.rs:749-756; a future refactor could re-introduce the hardcoded false with no failing test

**Evidence:**
```
grep -rn 'GpuBatch.*Mesh|mesh.*premultipl|premultipl.*mesh' module/helper/tilemap_renderer/tests/ returns no matches; webgl.rs:749-756 fixed GpuBatch::Mesh arm to read premultiplied from params.texture but no test covers this path
```

**Action:**
- [ ] Add a test that verifies GpuBatch::Mesh with a premultiplied texture calls apply_blend with premultiplied=true


---

## 📖 DOCUMENTATION

🟡 INTRODUCED — Comment at module/helper/tilemap_scene/src/compile/frame.rs:369-371 reads 'each VertexCorners object (e.g. a per-player region) can be coloured independently', implying TintBehaviour::Flat is VertexCorners-specific; it is now universal across all 6 layer types via layer_base_tint

**Evidence:**
```
frame.rs:369-371 comment says 'each VertexCorners object'; layer_base_tint is called at 9 sites in frame.rs: lines 372, 1004, 1039, 1156, 1306, 1393, 1426, 1521, 1542 — covering all layer types, not just VertexCorners
```

**Action:**
- [ ] Update the comment at line 369 to state that Flat tint is universal across all layer types via the shared layer_base_tint helper


---

FairlyBet:
@wanguadr Thanks for the review — processed all six items. Each was treated as a hypothesis and verified against the code before acting.

## Summary

| # | Item | Outcome | Commit |
|---|------|---------|--------|
| 1 | Build failure | ✅ Fixed | `750bfba5` |
| 2 | `ImageAsset` lacks `Default`/`non_exhaustive` | ❌ Rejected | — |
| 3 | `BlendMode::Multiply` ignores premultiplied | ⚠️ Code change rejected, doc clarified | `e4379a63` |
| 4 | `corner_source` no compile-time validation | ✅ Fixed | `1f5af26d` |
| 5 | No regression test for mesh-batch premultiplied | ⚠️ Test infeasible, mitigated | `dd43034e` |
| 6 | `frame.rs` Flat-tint comment | ✅ Fixed | `a21ea8cc` |

## Details

**1 — Build failure (FIXED, `750bfba5`).** Reproduced: `cargo build --workspace --all-features` failed with `E0063: missing field premultiplied` at `hexagonal_map/src/main.rs:187`. The new `premultiplied` field was added to `ImageAsset` but this example construction site was not updated (all in-crate sites already were). Added `premultiplied: false`. Workspace now builds clean.

**2 — `ImageAsset` Default/non_exhaustive (REJECTED).** Both suggestions conflict with the crate's design and neither would have prevented the breakage. `lib.rs` carries `#![allow(clippy::exhaustive_structs)]` — a deliberate opt-in to exhaustive public structs; `#[non_exhaustive]` would make the cross-crate struct-literal construction the examples/tests rely on a hard error everywhere. `impl Default` wouldn't help either: the example used a full literal (no `..Default::default()` spread), which must name every field regardless; `id` has no meaningful default and sibling asset structs impl none. The real defect was the single stale site fixed in item 1.

**3 — Multiply premultiplied (code change REJECTED, doc clarified, `e4379a63`).** The suggested swap of `DST_COLOR`→`src_a` would remove multiply entirely — `DST_COLOR` is the defining source factor (`src.rgb·dst.rgb`). `premultiplied` only selects the alpha-compositing factor (ONE vs SRC_ALPHA), which Multiply doesn't use, so its absence is correct — no double-alpha darkening. In fact the arm is *exact* for premultiplied sources at any alpha (`src.rgb = rgb·a` ⇒ `dst·(rgb·a + 1−a)` = reference). Clarified the comment to say so, mirroring the existing Screen-arm note.

**4 — `corner_source` validation (FIXED, `1f5af26d`).** Added a `validate()` rule flagging any `VertexCorners.corner_source` naming no object's `global_layer` (the only values `tile_corner_id` can match). Chose a structural validation rule over the suggested runtime warning: it fits the existing collect-all-violations architecture, needs no logging dependency, and leaves the deliberate compile-path VOID fallback (and its test) untouched. Added positive + negative tests.

**5 — mesh-batch premultiplied test (infeasible, mitigated, `dd43034e`).** A literal test asserting `apply_blend(..., premultiplied=true)` needs a live `WebGlTexture` + WebGL context; the webgl adapter only compiles for wasm32 and the crate has no wasm-bindgen-test/headless-GL harness, so a host test could only ever observe the `false` branch — useless against a hardcoded-`false` regression. Best available mitigation: extracted the duplicated resolution (single-mesh + mesh-batch paths) into one `GpuResources::mesh_premultiplied` helper, so the regression surface is a single shared function.

**6 — Flat-tint comment (FIXED, `a21ea8cc`).** Verified `layer_base_tint` is invoked from 10 sites across every layer type. Reworded the comment to state Flat tint is universal, not VertexCorners-specific.

All commits build clean (`--workspace --all-features`) and the `tilemap_scene` / `tilemap_renderer` test suites pass.

